PR #27, opened 2019-01-08
Added ro translations

realProngsDark:


alainm23:
Thank you very much for the translation, I have no idea because Travis marks it as failed, I will review it tomorrow at the first hour.

PR #46, opened 2019-01-19
Update French translations

NathanBnm:
Notes: 
In `fr.po`
- There is an error at line 817, it's written "proyect" instead of "project". I made the change in fr.po, I mean you have to correct for others languages.
- Same at line 924 it's written "goqnals" instead of "goals".
- Lines 1028-1029 are in Spanish instead of English


alainm23:
Hey, thank you very much for the translation and its corrections, I'm going to fix this.

PR #48, opened 2019-01-24
Fix some translations

NathanBnm:
I corrected some translations after seeing them in their context in the app. Despite, I discovered some errors. Here is the detail:

I based myself on the corrected version of the `fr.po` file for the line numbers.

- Strings at lines `262`, `509`, `596`, `610`, `625` and `661` aren't translated in the app when they are in the file.
- In the "menu" icon, "Light background" and "Dark background" aren't in translations files so they can't be translated for now.
- It's the same for "Reminder" when you add a task.
- And still the same for "Import" in the three dot menu next to the "Add a project" button.

There is also a problem with "AM" and "PM", please let people choose between 12 and 24 hour format. In French "AM" and "PM" doesn't make any sense. So maybe you have to look there.

Finally can you explain the role of the "Inbox" tab, because in French it's translated as "Boîte de réception" which is usually used for email inbox.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -86,11 +86,11 @@ msgstr "%A %d %b"
 
 #: src/Views/Inbox.vala:43
 msgid "All clear"
-msgstr "Tout est libre"
+msgstr "Il n'y a rien ici"
 
 #: src/Views/Inbox.vala:44
 msgid "Looks like everything's is organized. Tap + to add a task."
-msgstr "On dirait que tout est organisé. Tapez sur + pour ajouter une tâche."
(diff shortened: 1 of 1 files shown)


alainm23:
Thanks for the translations:
- For 'Inbox', maybe it could only be 'Inbox'
- I will be very aware of adding the 12 and 24 hour format.

NathanBnm:
> 
>     * For 'Inbox', maybe it could only be 'Inbox'
>
Is Inbox the place where tasks are going when they need to be done? Maybe you have to make this role more clear.
> 
>     * I will be very aware of adding the 12 and 24 hour format.
>
Thanks for that

Please don't forget to fix missing strings :+1:

PR #50, opened 2019-01-29
Update lt.po

welaq:


alainm23:
Thank you very much for the translations. Just in time before a new update.

PR #53, opened 2019-02-03
Fix overflow

harisvsulaiman:
Possible fix for https://github.com/alainm23/planner/issues/52


alainm23:
I'm not sure about using a ScrollWindow, I'd like the notes to always be completely visible.

harisvsulaiman:
@alainm23 
Seems like there isnt a max width attribute for gtk css.
And the other way was to set size on size change signal.

Scrolled window ensures that we can set a width.
Otherwise the issues that are loaded from GitHub tend to make the whole ui overflow.

harisvsulaiman:
You can remove max height on scrolled window to make the notes visible in full height.
And maybe set a horizontal scroll policy if needed

PR #72, opened 2019-06-18
Update German translation

ghost:
Signed-off-by: Malte Kiefer <malte.kiefer@kiefer-networks.de
In the next days, I will fix the rest.


alainm23:
Thank you very much for the translation, I will send an update before this weekend.

PR #82, opened 2019-10-03
Update french translation

dabou:
Update fr.po file.


alainm23:
I made some changes here, Can you update your PR pls?

dabou:
There is so much difference that I will do a new clean PR after updating my fork.

alainm23:
Ty very much.

PR #108, opened 2019-12-09
Update french translation

dabou:
Update french translation with the new updated po file.


alainm23:
Great. I'm going to check it when I coming to home.

PR #123, opened 2020-01-06
Add german translation 

thegesuser:
Like proposed on twitter, I localized the "de.po" file and created a pull request


alainm23:
Great, everything is correct, thank you very much 🙌️.

PR #194, opened 2020-02-01
Use proper version in preferences

### LINKED ISSUE #190: "About" should always show the latest version
Even after the update to 2.1.1 it still says "v2.0.8"
I wonder: Can Planner get the current version in "About" automatically or do you have to change it manually for each new version?

worldofpeace:
Before it was done manually when we have this
in Constants.

Fixes https://github.com/alainm23/planner/issues/190.


alainm23:
Great!!!

PR #212, opened 2020-02-09
Update es.po

drieiro:


alainm23:
Travis CLI stopped working, but everything is fine, thank you very much.

PR #277, opened 2020-03-11
Translation Portuguese (Portugal) pt_PT.po

rottenpants466:
Renamed pt.po to pt_PT.po.

In this way we have Portuguese (Portugal) and the existing Portuguese (Brazil) translations.

### DIFF
--- a/po/pt_PT.po
+++ b/po/pt_PT.po
@@ -9,17 +9,17 @@ msgstr ""
 "Project-Id-Version: com.github.alainm23.planner\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2020-03-02 18:53-0500\n"
-"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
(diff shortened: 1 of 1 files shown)


rottenpants466:
Uh could some one help me out here. I dont know why the check failed :|

This is my first time contributing something so maybe i did something wrong?

alainm23:
Thank you very much for helping to translate Planner. The error that Travis CLI gives is because the pt.po file does not exist. You should remove this line to fix it.

https://github.com/alainm23/planner/blob/master/po/LINGUAS#L136

However, the most recommended would be to add a new language.

rottenpants466:
Uh could some one help me out here. I dont know why the check failed :|

This is my first time contributing something so maybe i did something wrong?

rottenpants466:
Ok thanks. Should i close this pull request and make another one?

alainm23:
Well, it depends on what you want to do. If you want to remove pt.po and replace it with pt_PT.po just remove that line above and add pt_PT. But the most recommended would be to add a new language, so for this you would have to close this PR and open a new one.

Issue #126, opened 2020-01-09
Building from source fails on GNOME 3.34.3


yochananmarqos:
**Describe the bug**
Build fails with:

```
Compilation failed: 45 error(s), 3 warning(s)
[6/66] Generating com.github.alainm23.planner.appdata.xml_data_merge with a custom command.
ninja: build stopped: subcommand failed.
```

**To Reproduce**
Attempt to build on an Arch-based distro

**Expected behavior**
Self-explanatory

**Desktop**
 - Manjaro
- GNOME 3.34.3

**Additional context**
I ran into the same issue as in #119 as `evolution-data-server` 3.34.3-1 provides `libecal-2.0`. I bumped the version in `meson.build` to no avail. 

I know you're building this on and for Elementary OS, but it would be nice if it at least worked on The DE it's based on as well.

[Build log](https://del.dog/yupicyrett)

alainm23:
Could you rebuild please. I just updated the support for `libecal-2.0`

yochananmarqos:
The build still fails. 

[Build log](https://del.dog/firasapafe)

alainm23:
One more time pls. We go from 45 errors to 8.

yochananmarqos:
Success! 

The dark theme needs a little work. I'm using Adwaita-maia-dark:

![Screenshot from 2020-01-09 09-04-47](https://user-images.githubusercontent.com/5437803/72083486-250d6a80-32bf-11ea-8fc6-eec8e3c7de82.png)

alainm23:
Are you building Planner using flatpak?

yochananmarqos:
No, I'm building from source. See my PKGBUILD in the first post.

alainm23:
Understood, I am programming Planner from elementary OS, for other distributions I recommend using flatpak.

```
sudo apt install flatpak-builder
flatpak install flathub org.gnome.Sdk/x86_64/3.34
flatpak install flathub org.gnome.Platform/x86_64/3.34
flatpak install flathub io.elementary.BaseApp/x86_64/juno-19.08

git clone https://github.com/alainm23/planner.git
cd planner/build-aux/flatpak

flatpak-builder --install repo com.github.alainm23.planner.json --force-clean --user
```

Try this and Planner works better.

yochananmarqos:
Arch packages are built from source. I don't use flatpaks or snaps.

alainm23:
[...]

PR #290, opened 2020-03-31
Add Fedora installation instructions

tim77:
Hello. Packaged and available in official Fedora repos. Now [on QA](https://bodhi.fedoraproject.org/updates/FEDORA-2020-58dad2a040).

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,9 @@
 ### Arch Linux
 [elementary-planner-git](https://aur.archlinux.org/packages/elementary-planner-git) are available in AUR thanks to @yochananmarqos.
 
+### Fedora
+[elementary-planner](https://src.fedoraproject.org/rpms/elementary-planner) are available in Fedora repos thanks to @tim77.
+
 ## Building
 
 You'll need the following dependencies:


alainm23:
Thank you for the contribution. It is good to know that Planner will be available in more distributions.
I only have one question. What will happen when Planner has a new version available?

tim77:
> What will happen when Planner has a new version available?

My packages is always up to day. 😊 Fedora have Anitya monitoring service which notify on every new release.

alainm23:
Thank you very much, I like that solution.

gokberkakdeniz:
I think it is better to add the following dependencies as added to the Arch linux package:
- elementary-theme
- elementary-icon-theme

https://github.com/alainm23/planner/issues/126#issuecomment-573092461

tim77:
> I think it is better to add the following dependencies as added to the Arch linux package:
> 
>     * elementary-theme
> 
>     * elementary-icon-theme
> 
> 
> [#126 (comment)](https://github.com/alainm23/planner/issues/126#issuecomment-573092461)

Fedora is not Arch. Fedora have guidelines for such things. And there is no requirements or recommendation for adding 3rd-party GTK themes to apps and Elementary apps in particular. If you think Elementary apps should use elementary theme or should requires elementary themes package please discuss this this in Fedora devel mailing list.

gokberkakdeniz:
> > I think it is better to add the following dependencies as added to the Arch linux package:
> > ```
> > * elementary-theme
> > 
> > * elementary-icon-theme
> > ```
> > 
> > 
> > [#126 (comment)](https://github.com/alainm23/planner/issues/126#issuecomment-573092461)
> 
> Fedora is not Arch. Fedora have guidelines for such things. And there is no requirements or recommendation for adding 3rd-party GTK themes to apps and Elementary apps in particular. If you think Elementary apps should use elementary theme or should requires elementary themes package please discuss this this in Fedora devel mailing list.

Hmm, I don't know Fedora guidelines but i said that because unfortunately the app become useless without those themes since its ui become broken. Anyway, thank you for packing this app.

tim77:
Alsmost all Elementary apps useless with default Adwaita GTK theme. I 100% agree there. But i can't do much there, so that’s why i recommenced to talk about this with community. I am interesting in this discussion too.

alainm23:
@tncga Thank you very much for the observation, Planner is configured to use the default elementary OS theme and icons so it is necessary that these two packages are installed.

If these two dependencies cannot be added to Planner's rpm package I think we can tell the user to install it manually.

It is not necessary for the user to select elementary OS as their main theme, it is enough that the theme is installed for Planner to find and use it.

@tim77 Do you think this is possible?
@tncga Does installing these packages solve the problem?

[elementary-icon-theme]( https://src.fedoraproject.org/rpms/elementary-icon-theme)
[elementary-theme](https://src.fedoraproject.org/rpms/elementary-theme)

Please help me test this, I am designing a new website for Planner and would like to be able to add a more complete installation manual.

gokberkakdeniz:
> @tncga Thank you very much for the observation, Planner is configured to use the default elementary OS theme and icons so it is necessary that these two packages are installed.
> 
> If these two dependencies cannot be added to Planner's rpm package I think we can tell the user to install it manually.
> 
> It is not necessary for the user to select elementary OS as their main theme, it is enough that the theme is installed for Planner to find and use it.
> 
> @tim77 Do you think this is possible?
> @tncga Does installing these packages solve the problem?
> 
> [elementary-icon-theme](https://src.fedoraproject.org/rpms/elementary-icon-theme)
> [elementary-theme](https://src.fedoraproject.org/rpms/elementary-theme)

Yes it solves the problem. Before installation it was looking like [this](https://github.com/alainm23/planner/issues/126#issuecomment-572629409).
 Huge black border outside the main window still exist but it is on other DEs/distros too.

Issue #160, opened 2020-01-25
The notes part of a task should handle links to websites


4jNsY6fCVqZv:

4jNsY6fCVqZv:
What kind of help would you like to see for this issue?

alainm23:
I would like to know how do it. I haven't found sample code that does this functionality,

4jNsY6fCVqZv:
Meaning you've thought about it, too?
Do you think popular notes apps or markdown editors located in the AppCenter could be a good inspiration?

alainm23:
Yes, using Markdown is a good solution. However, in order to view a file in Markdown it is necessary to create a web view and I feel that this is a very complicated process.

4jNsY6fCVqZv:
Does Tasks also work with web views?
https://github.com/elementary/tasks

romixch:
Couldn't we just use a Gtk.Label formatted with Pango? That would enable us to render links. Most of the Markdown syntax could be mapped to pango. We could just render a label in front of the Gtk.Textview. When someone clicks onto the label, we quickly change to edit mode and render the Textview. Trello implements it the same way.
I would like to implement this as I find it very useful too. But let's start with just the link and without the whole markdown.

alainm23:
Something like this?

![image](https://user-images.githubusercontent.com/33765137/79163525-ddf2b200-7da4-11ea-8f93-c1aa502879a9.png)

![image](https://user-images.githubusercontent.com/33765137/79170770-fae3b100-7db5-11ea-81c8-987643c0edec.png)

romixch:
Looks very promising. Thanks a lot!
I also had a try yesterday evening. I think there are some corner cases in the function get_markup_format (). I will soon send you a pull request to make this even better.

4jNsY6fCVqZv:
Why is it that the link is only recognizable as such if the opened task note has no focus? As soon as I focus the note section of the task that contains the note, the URL formatting disappears again and the link turns black. Until I move the focus away again.

Is there a solution for a persistent rendering?
I have also noticed that the project description can't handle links. Why is there a difference to the task note here?

alainm23:
Because it is a solution that other task managers use as @romixch says, I like the idea of ​​rendering in real time but I'm not sure if Gtk supports it, but for me it works very well.

romixch:
I think it is better not to handle links when the field is in edit mode. It may otherwise be annoying when you want to change the url, click on it and the browser pops up. 

[...]

PR #309, opened 2020-04-15
Fix Bug #160: Better url detection

romixch:
URLs that contained an ampersand (&) did not work. Made it entirely with regex matching.


alainm23:
Thank you very much. You improved the process.

PR #316, opened 2020-04-19
Improve appearance when built like explained in README.md

romixch:
I added a condition to the point where the theme elementary is set. It only sets the theme when it can be found on the system. This way Planner takes on the local gtk theme as a fallback. That makes Planner look more adapted to the local environment and does not show a broken UI when the theme elementary can't be loaded.

On my machine this results in this:
grafik

Instead of this:
grafik

I tested the flatpak build. It is not affected by this change.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -128,8 +128,12 @@ public class Planner : Gtk.Application {
         utils.apply_theme_changed ();
 
         // Set Theme and Icon
-        Gtk.Settings.get_default ().set_property ("gtk-icon-theme-name", "elementary");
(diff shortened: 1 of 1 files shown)


alainm23:
Thank you very much for your PR. The beginning of the project we decided to enforce the elementary OS theme for many reasons.
The most obvious are:
- Ability to fully control the UI and UX of the application on every platform.
- Offer a consistent user experience to everyone.
- Remove the burden of supporting a million different DEs and Themes.
- Keep the focus on the architecture instead of spending time fixing themes issues.

We recommend installing Planner from AppCenter and Flathub and avoid building from the repository.

PR #339, opened 2020-05-01
Database signal to reload data

romixch:
Hi Alain

Hope you are ok.

This pull request introduces a signal to reload the data. It will help implement a database import as well as the use of a different database file.

The two background services *Notification* and *Todoist* will stop and restart as well.

I already tested this with my other branch for changing the database location. Works pretty well.

Stay healthy

Roman


alainm23:
Many thanks. Definitely you will improve the process, I will do some tests before doing merge.

PR #358, opened 2020-05-09
Add Arc-Dark theme

Hanouta:
This adds a theme option which is based on the Arc-Dark theme. It's a pretty popular theme, especially useful for people who don't want to use a light theme but find the other Dark theme options to be too dark.

### DIFF
--- a/src/Utils.vala
+++ b/src/Utils.vala
@@ -741,6 +741,17 @@ public class Utils : GLib.Object {
                 row_selected_color = "shade (#ffffff, 0.125)";
 
                 Gtk.Settings.get_default ().gtk_application_prefer_dark_theme = true;
+            } else if (appearance_mode == 3) {
+                base_color = "#353945";
+                check_border_color = "grey";
+                projectview_color = "#404552";
+                pane_color = "#353945";
+                pane_selected_color = "#2B303B";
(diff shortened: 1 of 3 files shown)


alainm23:
I like the theme.

image

src/Utils.vala
```diff
@@ -740,6 +740,17 @@ public class Utils : GLib.Object {
...
+                Gtk.Settings.get_default ().gtk_application_prefer_dark_theme = true;
+            } else if (appearance_mode == 3) {
+                base_color = "#353945";
+                check_border_color = "grey";
+                projectview_color = "#404552";
+                pane_color = "#353945";
+                pane_selected_color = "#2B303B";
+                pane_text_color = "#fefeff";
+                popover_background = "#353945";
+                row_selected_color = "shade (#fefeff, 0.25)";
```
alainm23: I think we should improve the border color of the task, maybe a deeper black.
image

alainm23 (changes requested):

src/Utils.vala
Hanouta: Just pushed an update, do you like the new one better?

alainm23:
Thank you very much, it looks better now.

PR #365, opened 2020-05-12
fix #363: Make email addresses clickable

### LINKED ISSUE #363: Task notes should be able to handle mailtos
https://github.com/alainm23/planner/issues/160#issuecomment-626839637

romixch:


alainm23:
Thank you very much, super simple but functional.

PR #373, opened 2020-05-15
Changed item spacing in pane and bold only for selected item

ni-ka:
planner-item-spacing

### DIFF
--- a/data/stylesheet.css
+++ b/data/stylesheet.css
@@ -169,6 +169,11 @@ paned.horizontal > separator {
 */
 
 .pane-item {
+    padding-top:.2em;
+    padding-bottom:.2em;
+}
+    
+.pane-row:selected .pane-item {
     font-weight: 600;
 }
 


alainm23:
Thanks. Why did you decide to change the space between each item? I'm not sure to apply bold to a selected item because folders already use a bold label.

ni-ka:
Sorry my description should have been clearer. I found the current style (everything bold) a bit straining on the eye, so I removed the bold on all items and replaced with selected only. The spacing look better to me, also took inspiration from the todoist UI.

Check a comparison (from left to right: planner master, planner this branch, todoist):
sidebar-styles

alainm23:
Thank you very much.

4jNsY6fCVqZv:
> I found the current style (everything bold) a bit straining on the eye

Hello, I don't understand the issue yet. I use the native version 2.3.5 for from the elementary AppCenter. Also with the native button layout. In both the Dark Theme and the Light Theme, only the views and folder names are bold. Everything else is not.

Issue #367, opened 2020-05-14
Support using tags and labels in quick-add (ex today, @label1, #my-project, and p1 for priority)


ni-ka:
**Is your feature request related to a problem? Please describe.**
When I use the todoist web app, I can directly add a task to tomorrow by using "Call Amanda tomorrow #my-project". I can also assign to a project without using the mouse to choose a project

**Describe the solution you'd like**
It would be great if tags could be assigned rather than interpreted as part of the task title, to mirror the todoist web app behavior. Ideally it could offer autocomplete, but at first it is not necessary. It should support scheduling (today, tomorrow), tags @, labels # and priorities.

**Describe alternatives you've considered**

**Additional context**

benmordecai:
Its been a long time since this has been looked at. Are there plans to put todoist style incline scheduling, #projects, @tags, and !!1 priorities?

alainm23:
I have a small preview of this feature
https://github.com/user-attachments/assets/603675ef-a77e-4167-acae-5e5a1927a952

alainm23:
I have a small preview of this feature

https://github.com/user-attachments/assets/7ea3bcdc-9dc7-44df-b5e0-0f4796507cf9

benmordecai:
Looks perfect!

PR #374, opened 2020-05-16
Added ability to use tags in new item text: today, tomorrow, p1 to p4

ni-ka:
First attempt at addressing #367. Only in app so far, not in quick-add since it uses separate code.

### DIFF
--- a/src/Utils.vala
+++ b/src/Utils.vala
@@ -25,6 +25,9 @@ public class Utils : GLib.Object {
     private const string ALPHA_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
     private const string NUMERIC_CHARS = "0123456789";
 
+    private const string TODAY = _("today");
+    private const string TOMORROW = _("tomorrow");
+
     public string APP_FOLDER; // vala-lint=naming-convention
     public string AVATARS_FOLDER; // vala-lint=naming-convention
     public Settings h24_settings;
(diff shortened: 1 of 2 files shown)


src/Utils.vala
```diff
@@ -1016,4 +1016,42 @@ public class Utils : GLib.Object {
...
+        Regex wordRegex = /\S+\s*/;
+        MatchInfo matchInfo;
+        
+        var matchText = text. strip ();
+        for (wordRegex.match (matchText, 0, out matchInfo) ; matchInfo.matches () ; matchInfo.next ()) {
+            var word = matchInfo.fetch (0);
+            var stripped =    word.strip ().down ();
+
+            switch (stripped) {
+                case "today":
```
alainm23: Use `_("today")` to make the text translatable

src/Utils.vala
```diff
@@ -1016,4 +1016,42 @@ public class Utils : GLib.Object {
...
+                    item.due_date = new GLib.DateTime.now_local ().to_string ();
+                    break;
+                case "tomorrow":
```
alainm23: Use `_("tomorrow")` to make the text translatable

alainm23 (changes requested):
Thank you very much, this is a great first step. I just have a few observations. Later I would like to make it more visual.

src/Utils.vala
ni-ka: will do!

src/Utils.vala
ni-ka: will do!

ni-ka:
I've added the translation texts as constants as it would not let me use _() in the case statement.

> Later I would like to make it more visual.
Yeah, I've started playing around with replacing the entry text with a GtkSourceTextView to have completion (for keywords, projects and labels) and being able to add visual tags, but it'll probably take me a bit.

PR #385, opened 2020-05-22
Fix gtk criticals

romixch:
I wanted to get rid of Gtk-Critical warnings in the log. This are three small changes that make the log in while development much more readable.


alainm23:
Thank you very much, how did you know which widget gave the gtk criticals?

romixch:
I first tried to find it with setting G_DEBUG=fatal-criticals in order to get a stack trace. But that didn't help me much. 
Then I disabled different areas of the UI to see if the critical still appears. I found out that it was in the left pane. So I narrowed it down with the same strategy until I ended up at that control.
Was fun to debug and learn about GLib, Vala and coredumpctl.

alainm23:
Thank you very much for your time, keep it up ☺️.

PR #387, opened 2020-05-23
Update ro.po

kepler-dslr:
I'm a Romanian and I'll love this program in Romanian too :)


alainm23:
Thank you very much, everything perfect. But I would recommend using our Weblate service to suggest new translations. https://hosted.weblate.org/projects/planner/

kepler-dslr:
I didn't know that the project have a weblate. Good to know

PR #481, opened 2020-08-01
Fix search for single quotes: use of prepared statements

yannicka:
Before this PR, if I'm searching for `'`:

```
ERROR:../src/Services/Database.vala:1684:services_database_get_all_projects_by_search: assertion failed: (res == Sqlite.OK)
Bail out! ERROR:../src/Services/Database.vala:1684:services_database_get_all_projects_by_search: assertion failed: (res == Sqlite.OK)
fish : Tâche , 'com.github.alainm23.planner' terminée par le signal SIGABRT (Abandon)
```

This PR fix that.


alainm23:
Thank you

PR #484, opened 2020-08-02
Avoid unnecessary repetition when retrieving items from the database

yannicka:
Avoid unnecessary repetition when retrieving items from the database. It is probably possible to do even better later.


alainm23:
Thank you very much, this is a great way to optimize the code.

PR #486, opened 2020-08-03
Find in notes

### LINKED ISSUE #475: Ability to search in task notes too
When preforming search, it doesn't look for text in tasks content (with the exception for subtasks text content).
It would be nice to search in all content :)

Thank you for your app so far, really good.

yannicka:
Fix #475.

Peek-03-08-2020-21-01

I didn't work on the design, don't hesitate to improve.


alainm23:
Thank you very much for your implementation.

PR #521, opened 2020-09-20
Update README.md

Brin-o:
Updated the dimensions of the image at the very start of the file to make it not look squished.


alainm23:
Thanks!!!

Issue #532, opened 2020-09-25
Fearure request: Hide all details (collapse all)


alexios-tsiaparas:
When I am reviewing the tasks for the day, I normally open each one in order to see any notes. When finished ordering them, I would like to collapse all of them, so that more can fit in my screen. This seems to be impossible currently.

We could add a keyboard shortcut (I personally prefer them) to accomplish that, and maybe even an icon next to the view (Inbox, today, upcoming, project), like the hide details button on each task.

alainm23:
I'm using a keyboard shortcut for testing: `Shift + h` but this is yet to be defined.

PR #535, opened 2020-09-27
Hide all details (collapse all)

alexios-tsiaparas:
This pull request corresponds to my feature request #532. I implemented a button that collapses all the tasks in a view. I placed it after the name and other parameters at the left of the header.

image

I will be implementing  shortcut key too, just needed this for every day, so I was in a hurry.

Button placed on each view that collapses all tasks

Signed-off-by: Alexios Tsiaparas <alexios@watergate.gr>

Hope you accept it

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -246,9 +246,9 @@ public class Planner : Gtk.Application {
         //  }
 
         //  if (version) {
-		//  	print ("Test 0.1\n");
-		//  	return 0;
-		//  }
+        //      print ("Test 0.1\n");
+        //      return 0;
+        //  }
 
         return app.run (args);
     }
--- a/src/Views/Completed.vala
+++ b/src/Views/Completed.vala
@@ -36,6 +36,13 @@ public class Views.Completed : Gtk.EventBox {
         title_label.get_style_context ().add_class ("title-label");
         title_label.use_markup = true;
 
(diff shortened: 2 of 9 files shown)


alainm23:
Thanks for working on this feature, I had planned to work on the feature before the new update, but your work is a good start.
Issues to consider:
- Using a button is a good solution, but I'm not sure if the position where it is placed is correct, I like the idea of ​​using a specific keyboard shortcut for this feature better, what do you think and what keyboard shortcut could be assigned?
- Your solution is correct but it only works with independent tasks, the tasks within a section are not hidden, we could implement a global event for this.

alexios-tsiaparas:
If you have any preferences regarding the position, please share them with me and I can place it. I use the button a lot when planning, so I prefer it being next to the title, it makes it easy to access. We could also place it on the right side, with the rest of the buttons. Other than that, I cannot think of a suitable position.

I would propose Shift+c, from close, but Shift+h from hide is equally fine I guess. If you are OK, I will implement it with Shift+h that you use and if we decide otherwise it should be easy to change at this stage.

Sorry for the sections, but I have not yet used them and totally forgot about them. I will implement them as well.

alexios-tsiaparas:
I think you have implemented the sections, is that so? I see that there are conflicts, will you resolve them or should I do that. In the files project.vala and inbox.vala, is your code replacing mine or is it meant for the sections?

PR #576, opened 2020-10-23
Update it.po

albanobattistella:


alainm23:
Thank you very much for your translation, unfortunately I have to decline because we do not support direct translations in the repository, use our translation service in [Weblate](https://hosted.weblate.org/projects/planner/translations/) for this purpose.

albanobattistella:
> Thank you very much for your translation, unfortunately I have to decline because we do not support direct translations in the repository, use our translation service in [Weblate](https://hosted.weblate.org/projects/planner/translations/) for this purpose.

HI!,
done!
V/R,
Albano

PR #619, opened 2020-12-07
Add option to use system title bar

### LINKED ISSUE #406: [Feature Request] Title Bar support
It's not exactly a problem, but I use GNOME and every app that I uses has support to the system titlebar, giving a more cohesive look. I would love to see support for this in Planner app.

image

dsprenkels:
This PR adds the setting 'use-system-decoration', which will
disable the custom title bar that is used by default.

To enable/disable the `window-buttons` field in the preferences dialog, I had to add a method `set_combobox_sensitive`, which is used to mark the combobox as sensitive (i.e. enable/disable editing).

I could not find how to run the linter. I did my best to follow the style that was used; pardon me if there I missed something.

Fixes #406

### DIFF
--- a/src/Dialogs/Preferences/Preferences.vala
+++ b/src/Dialogs/Preferences/Preferences.vala
@@ -597,12 +597,23 @@ public class Dialogs.Preferences.Preferences : Gtk.Dialog {
         list.append ("Close Only Left");
         list.append ("Close Only Right");
 
+        var use_system_decoration_val = Planner.settings.get_boolean ("use-system-decoration");
         var button_layout = new Dialogs.Preferences.ItemSelect (
             _("Button layout"),
             Planner.settings.get_enum ("button-layout"),
             list
         );
+        button_layout.set_combobox_sensitive (!use_system_decoration_val);
         button_layout.margin_top = 12;
+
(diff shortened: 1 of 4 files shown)


Toxblh (approved):
So looks good to me. 
Also, I tried it and it works as expected
image

src/Dialogs/Preferences/Preferences.vala
```diff
@@ -597,12 +597,23 @@ public class Dialogs.Preferences.Preferences : Gtk.Dialog {
...
             _("Button layout"),
             Planner.settings.get_enum ("button-layout"),
             list
         );
+        button_layout.set_combobox_sensitive (!use_system_decoration_val);
         button_layout.margin_top = 12;
+
+        var use_system_decoration = new Dialogs.Preferences.ItemSwitch (
+            // The reason we need a restart is because GTK has to tell the
+            // window manager whether to add window deocartions *before* the
```
dsprenkels: There's a typo here in `deocartions`.

alainm23:
@dsprenkels Thank you very much for this feature, it looks a little weird but if it is what you need to use more Planner for me, no problem, I'll definitely make some changes and design improvements but for everything else it is fine.

PR #922, opened 2022-07-20
Fixed spelling error in PriorityButton

peterrauscher:
When choosing priority for a task, the Priority 1 option read "hight" instead of "high". This commit fixes that.

Thank you for this incredible project, I use it daily!


ghost:
Thanks, was just about to make a PR about this!

alainm23:
Fixed in planner-gnome branch

PR #968, opened 2023-01-29
application: Fetch translations from the weblate

ghost:
We don't want to lost all translations.


alainm23:
Thank you very much, but this is not necessary, I prefer to keep only the translation files available and not blank, until someone wants to create it. Task Planner has a good manual to do this.

https://github.com/alainm23/planner/tree/master/po

PR #985, opened 2023-05-10
build: Use gtk4-update-icon-cache

bobby285271:
Otherwise the post install script fails if gtk3 is not installed, which can happen in builder sandbox.


bobby285271:
The CI failures are not related to this PR I think.

alainm23:
Ty.

PR #919, opened 2022-07-03
Add Turkish language support

mucahitkurtlar:
I added Turkish language support to it.

PR #996, opened 2023-06-02
Add Turkish translation

ghost:
Based on Wablate, Improved over #919


alainm23:
Thanks for the translation

PR #997, opened 2023-06-02
Fix a typo in App.vala

ghost:


alainm23:
Thanks, just a small detail.

PR #1003, opened 2023-06-02
Mark three strings translatable

ghost:
- Mark three strings translatable
- Update Turkish translation


alainm23:
Thank you so much

PR #1004, opened 2023-06-02
fix icon in README.md

Drsheppard01:


alainm23:
A small detail, anyway we are still in search of a new icon. Thank you.

PR #1014, opened 2023-06-11
Updated French translation

rene-coty:
I updated the fr.po file
+ I noticed that the app's description isn't included in the translatable strings/.pot file, it would be great to add it…


alainm23:
Thank you so much, I'll keep it in mind

PR #1017, opened 2023-06-25
Updated French translation

rene-coty:
I updated the fr.po file + deleted the old .pot file wich lacked entries


rene-coty:
Btw I noticed that it is currently not possible to translate the content of the Shortcuts window. Could you add it to the translatable strings for next release ?

alainm23:
Thanks.

PR #1018, opened 2023-06-26
Add Simplified Chinese translation

Fr4nk1inCs:
Some translation is based on the original Planner Weblate translation.


alainm23:
Thanks @Fr4nk1in-USTC.

PR #1027, opened 2023-06-27
Correct tutorial misspelling

jntdofe:


alainm23:
Thanks for the request, but the corrections at the level of the translation file are not valid, keep in mind that this is generated every time an update is released.

The change would have to be done at the code level, as all translations will be updated.
https://github.com/alainm23/planify/blob/master/src/Util.vala#L1168

If you don't have time, no problem, I'll do it later, thanks for everything, I'll close this problem but I'll take it into account when updating the translations, thanks.

ghost:
Hello @alainm23 I still see this typo in the app - would it be okay to try my hand at filing a pull request as per your instructions here?

alainm23:
@evacks  I'm not checking typos in detail (English is not my native language), if you could help me with that I would be very grateful.

PR #1043, opened 2023-07-14
Fix some criticals, warnings and deprecations

Diego-Ivan:
Hello there! I've been using Planify for a couple of weeks and I've loved it. I made some small changes in the development version fixing some criticals that popped up as well as some warnings in compile time. These include:

* Remove deprecations in GTK 4.10 and Libadwaita 1.3 (latest in GNOME 44 runtime):
  * Use Gtk.DropDown instead of Gtk.ComboBoxText
  * Use a Gtk.Image prefix instead of Adw.ActionRow.icon_name
* Avoid illegal casts and calling methods from null instances
* Transfer ownership for delegate methods
* Fix an accelerator in the shortcut window
* Tweaks to the Flatpak Manifests

If there is any change I have to make to improve this PR I will make the necessary updates as soon as I can :smile:

### DIFF
--- a/build-aux/io.github.alainm23.planify.Devel.json
+++ b/build-aux/io.github.alainm23.planify.Devel.json
@@ -4,6 +4,9 @@
     "runtime-version": "44",
     "sdk" : "org.gnome.Sdk",
     "command" : "io.github.alainm23.planify",
+    "sdk-extensions" : [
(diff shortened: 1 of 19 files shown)


alainm23:
@Diego-Ivan  thank you very much for the PR, there are many things that I had not noticed and this solves it.
I invite you to continue making these types of reviews, thank you very much.

PR #1072, opened 2023-12-11
Updated French translation

rene-coty:
I updated the fr.po file


alainm23:
Thank you very much for the translations.

PR #1074, opened 2023-12-12
Adding Russian language

hachikoharuno:
Hello, this is my first pull request, so I’m not sure if I’m doing everything correctly, I just repeated after rene-coty, who did the translation into French.

I really like Planify and decided to translate it into Russian. I compiled the program myself and checked the translation so that everything was displayed correctly.


alainm23:
Thank you very much, I'll check it out

PR #1079, opened 2023-12-16
fix typos in welcome project

Ben9986:
There are two small typos in the "Meet Planify" project. This commit fixes them :)

### DIFF
--- a/src/Util.vala
+++ b/src/Util.vala
@@ -1164,7 +1164,7 @@ public class Util : GLib.Object {
             item_04.id = Util.get_default ().generate_id (item_04);
             item_04.project_id = project.id;
             item_04.content = _("Reorder yours to-dos");
-            item_04.description = _("To reorder your list, tap amd hold a to-do, then drag it to where it shpuld go.");
+            item_04.description = _("To reorder your list, tap and hold a to-do, then drag it to where it should go.");
 
             var item_05 = new Objects.Item ();
             item_05.id = Util.get_default ().generate_id (item_05);


alainm23:
Hello and thanks for the correction, the correct flow would be to first edit the source code, then generate the new translation files and fix the translation file if necessary.

1. Edit [this](https://github.com/alainm23/planify/blob/master/src/Util.vala#L1167) code line
2. Generate the new translation files using the command `ninja io.github.alainm23.planify-update-po`inside the build folder
3. Check if the translation files were updated correctly.

I understand that it is a long process, but just editing the source code file is enough. thanks.

po/io.github.alainm23.planify.pot
```diff
@@ -334,7 +334,7 @@ msgstr ""
 
 #: src/Util.vala:1165
 msgid ""
-"To reorder your list, tap amd hold a to-do, then drag it to where it shpuld "
+"To reorder your list, tap and hold a to-do, then drag it to where it should "
```
alainm23: Edit source code instead of translations

alainm23 (changes requested):

po/io.github.alainm23.planify.pot
Ben9986: sorry! unfamiliar with the project and admittedly didn't look hard enough for the correct file. I've edited the source file now

Ben9986:
I can't seem to get the build working on my setup. Can you check it and lmk if I've missed something

alainm23:
Don't worry, this is enough, thank you.

PR #1085, opened 2023-12-19
Update Russian translation

hachikoharuno:
Hello, I would like to clarify with you, I have added a translation for key combinations, which is not translated by default and the translation of which is not in the .pot file.

Do I need to remove the line data from the file?


alainm23:
Thank you for the translations. Regarding the translations of the keyboard shortcuts, I forgot to add it to the translation sources (Thanks for reminding me), the way you are adding it is invalid and will not work, these translation files are generated from the source code automatically. I would ask you not to wait until the translations are available.

po/ru.po
```diff
@@ -1556,182 +1470,217 @@ msgstr "Синхронизация с сервером"
...
 msgid "Planify - Quick Add"
-msgstr "Быстрое добавление"
+msgstr "Planify — Быстрое добавление"
 
-#~ msgid "Save"
-#~ msgstr "Сохранить"
 
-#~ msgid "AM"
-#~ msgstr "AM"
+#: shortcut
```
alainm23: This will not work, I will update the translations so you can add it.

alainm23 (changes requested):

alainm23:
Please resolve conflicts to continue, thank you very much.

alainm23:
Everything is correct, thank you very much.

PR #1092, opened 2023-12-25
Update ru translation

hachikoharuno:


alainm23:
Thank you very much for the translation, Planify is constantly updated and as a credit your Github user will appear in the next release.

PR #1103, opened 2024-01-13
Change strings to title case and remove periods

### LINKED ISSUE #1102: Text doesn't follow GNOME HIG
**Describe the bug**
The titles and descriptions in the app preferences use title case inconsistently, and descriptions use periods while they shouldn't, not following the GNOME HIG.

**To Reproduce**
1. Go to Preferences

**Expected behavior**
The text should be consistent.

**Screenshots**
image
image

 - OS or DE: GNOME
 - Version: 4.4

**Additional context**
I'll be pushing a PR later :)

jamiethecat:
Closes #1102

### DIFF
--- a/src/Dialogs/Label.vala
+++ b/src/Dialogs/Label.vala
@@ -92,7 +92,7 @@ public class Dialogs.Label : Adw.Window {
         color_group.add_css_class (Granite.STYLE_CLASS_CARD);
         color_group.attach (color_picker_row, 0, 0);
         
-        submit_button = new Widgets.LoadingButton.with_label (is_creating ? _("Add label") : _("Update label")) {
+        submit_button = new Widgets.LoadingButton.with_label (is_creating ? _("Add Label") : _("Update Label")) {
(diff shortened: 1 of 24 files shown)


src/Dialogs/Label.vala
```diff
@@ -218,3 +218,4 @@ public class Dialogs.Label : Adw.Window {
         });
     }
 }
+
```
TheEvilSkeleton:
```suggestion
```

TheEvilSkeleton (changes requested):

src/Dialogs/Label.vala
alainm23: According to [Vala Lint](https://github.com/vala-lang/vala-lint/wiki/Vala-Lint-Checks) there should be a blank line at the end. Although it is not a problem, we still need to improve the code for that Lint to pass.

alainm23:
Thank you very much, I will keep this format in mind for the future.

PR #1104, opened 2024-01-14
Update russain translate

hachikoharuno:
Hello, I would like to take this opportunity to recommend the boosty service (patreon in Russia), because all of the above support tools do not work in Russia due to sanctions.

Of course, I’m not sure how cost-effective this is, but if you don’t have any difficulty, you can try adding it to your links.


alainm23:
Thank you very much, the translations have been updated, could you please fix it.

PR #1108, opened 2024-01-14
po: Update Turkish translation

ghost:


alainm23:
Thank you very much, the translations have been updated, could you please fix it.

PR #1135, opened 2024-02-19
Various snap related changes

soumyaDghosh:
1. Updated the `quick-add` command when using as a snap
2. Updated gitignore to ignore snap related files
3. Updated README with snap release


alainm23:
Thank you very much.

soumyaDghosh:
I have kept the snap hidden in the store now, until we get an auto connection for a content snap which is needed for webkitgtk. Thank you for merging the PR.

PR #1208, opened 2024-03-24
Fix a crash if no color is defined for a Caldav Project

### LINKED ISSUE #1197: Crash with my CalDAV server/provider
**Describe the bug**

I use [murena.io](https://murena.io/) as my cloud provider, it is based on Nextcloud and uses CalDav to sync my calendar and tasks on my various devices/apps.
When I provide my credentials in Planify, the app will freeze and crash. If I open it again it will try to sync with my account and crash again, and again, making the app unusable unless I disconnect my laptop from Internet to stop the sync and deactivate CalDav integration in Planify.

**To Reproduce**

1. Go to 'Integrations'
2. Toggle 'CalDav' on
3. Type 'https://murena.io/' in the server entry field and provide your credentials
4. Click 'Connect' and wait → crash after a few seconds

**Expected behavior**

Planify should sync with my account without problem, it works on many other apps I have tested on Linux so far (Errands…). The problem doesn't come from my credentials, because Planify will display a proper error message if the URL or the password/username aren't correct.

**Screenshots**

[Capture vidéo du 2024-03-21 09-15-35.webm](https://github.com/alainm23/planify/assets/95506494/91f99980-5435-4437-b910-c93ab3ebc833)

 - OS or DE: GNOME
 - Version 45.5

byquanton:
Should fix #1197

### DIFF
--- a/core/Objects/Project.vala
+++ b/core/Objects/Project.vala
@@ -324,7 +324,11 @@ public class Objects.Project : Objects.BaseObject {
         GXml.DomElement propstat = element.get_elements_by_tag_name ("d:propstat").get_element (0);
         GXml.DomElement prop = propstat.get_elements_by_tag_name ("d:prop").get_element (0);
         name = get_content (prop.get_elements_by_tag_name ("d:displayname").get_element (0));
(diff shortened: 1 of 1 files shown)


alainm23:
Thank you very much,

Issue #1216, opened 2024-03-27
Task action button tooltips aren't translated (but are translatable?)


Gert-dev:
**Describe the bug**
The task action tooltips don't appear to be translated. I [can find some of them in the translation files](https://github.com/alainm23/planify/blob/master/po/nl.po#L594), though.

The priority dropdown items also aren't translated.

**To Reproduce**
1. Open the application in another locale (e.g. Dutch).
2. Hover over the task action buttons when expanding an existing task.
3. Observe them not being translated.
4. Click 'Set The Priority'
5. Observe the dropdown items also not being translated.

**Expected behavior**
The tooltips and dropdown items are translated.

**Screenshots**
![image](https://github.com/alainm23/planify/assets/24809863/a859c7b4-f224-4a3f-a96c-a94da3761513)

![image](https://github.com/alainm23/planify/assets/24809863/3a09dc68-4598-4f80-97d1-0cafb784b20a)

**Desktop (please complete the following information):**
 - OS or DE: GNOME
 - Version: 4.5.10

**Additional context**
The 'pin' button tooltip _is_ translated, strangely enough, it's just the four other actions that aren't:

* Add Subtasks
* Add Labels
* Set The Priority
* Add Reminders

... and also the priority dropdown tooltips also isn't translated, as mentioned.

alainm23:
The translation files were updated, you could generate a PR to update the translation.

PR #1222, opened 2024-03-28
Update Dutch translation

Gert-dev:
This updates the Dutch translation as requested in https://github.com/alainm23/planify/issues/1216#issuecomment-2024157856.


alainm23:
Thank you

PR #1271, opened 2024-05-01
Fix priority icon colors

Riksu9000:
Priority icons 2 and 3 used colors from icon 1.

Before
Screenshot from 2024-05-01 11-31-00
After
Screenshot from 2024-05-01 11-59-05


alainm23:
Thank you very much

PR #1272, opened 2024-05-01
Fix view option menu indicator covering the button

Riksu9000:
Previously clicking above the indicator would do nothing. Now clicking above it is the same as clicking the button.

The indicator doesn't seem to be used in ReminderButton and the Scheduled view, but I updated it regardless. Also the indicator is misaligned in the Today view, but that's unrelated.


alainm23:
Thank you very much, I hadn't noticed

PR #1279, opened 2024-05-05
Fix scheduled range view

### LINKED ISSUE #1270: Section "Scheduled": Area for "rest of month" (e.g. May 08-31) does not display scheduled events
**Describe the bug**
May is the upcoming month and today the 30st April. Therefore the section "Scheduled" shows 7 subsections for May 1st until 7th. These work properly. But the section May 08-31 does not. As soon as I schedule a task for this range, it is not displayed anymore.

This is also true for any detected appointment from calendars like Evolution, Thunderbird etc.

**To Reproduce**

Set a task’s schedule to pre May 8th or into June (or even later) and switch to Panel for Schedule: it is visible once again. Change that task’s schedule to sth. in the mentioned range: it "vanishes".

 - OS or DE: Ubuntu 22.04 (currently using x11 and not wayland … due reasons)
 - Version of planify is 4.6, according to "flatpak list" (io.github.alainm23.planify)

**Additional context? Nope, 'just' feedback**
Thanks for creating this nice app; hope it will be around for some time.

I finally decided to leave the unmaintained TaskCoach behind, which was a nice and powerful tool, but in the end, the Efforts feature had no real added value for me and was rather tedious (did only take several years to finally admit this one … *cough*)

Riksu9000:
Fixes #1270

In other scheduled views, listbox_revealer is appended, but here listbox_grid is appended instead, which is already a child of the revealer. I'm assuming it fails because a widget can't have two parents.

### DIFF
--- a/src/Views/Scheduled/ScheduledRange.vala
+++ b/src/Views/Scheduled/ScheduledRange.vala
@@ -122,7 +122,7 @@ public class Views.Scheduled.ScheduledRange : Gtk.ListBoxRow {
         });
         
         content.append (event_list_revealer);
-        content.append (listbox_grid);
+        content.append (listbox_revealer);
 
         child = content;
 


alainm23:
Thank you.

PR #1288, opened 2024-05-11
README: Refer to Planify

genodeftest:
Fix app name, replacing a name that probably has been copy-pasted


alainm23:
Thank you very much

PR #1290, opened 2024-05-12
Make filter icon animation smoother

Riksu9000:
With these changes, the animation won't stop abruptly.

Previously clicking on the filter multiple times or clicking on another filter while the animation is playing can cause the animation to end abruptly. Now the animation always plays to the end before it can start again.

The animation length is 700ms, so the timeout was made shorter.

### DIFF
--- a/src/Layouts/FilterPaneRow.vala
+++ b/src/Layouts/FilterPaneRow.vala
@@ -94,14 +94,16 @@ public class Layouts.FilterPaneRow : Gtk.FlowBoxChild {
         Services.EventBus.get_default ().pane_selected.connect ((pane_type, id) => {
             if (pane_type == PaneType.FILTER && filter_type.to_string () == id) {
                 add_css_class ("selected");
-                add_css_class ("animation");
-                Timeout.add (1000, () => {
-                    remove_css_class ("animation"); 
-                    return GLib.Source.REMOVE;
-                });
+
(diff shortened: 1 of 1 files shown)


src/Layouts/FilterPaneRow.vala
```diff
@@ -93,15 +93,20 @@ public class Layouts.FilterPaneRow : Gtk.FlowBoxChild {
 
         Services.EventBus.get_default ().pane_selected.connect ((pane_type, id) => {
             if (pane_type == PaneType.FILTER && filter_type.to_string () == id) {
+                var css_classes = get_css_classes ();
```
alainm23: You could use `has_css_class` instead of traversing the list of classes.
https://valadoc.org/gtk4/Gtk.Widget.has_css_class.html

alainm23 (changes requested):

src/Layouts/FilterPaneRow.vala
Riksu9000: Ah, that makes a lot of sense, thanks.

Riksu9000:
I noticed an issue where the selected style is not always applied. I'll take another look tomorrow.

Riksu9000:
Never mind, it was pretty simple.

alainm23 (approved):

PR #1292, opened 2024-05-12
Changed settings entry name to the actual behaviour

dryBoneMarrow:
When the setting "Underline Completed Tasks" is enabled, completed tasks are crossed out.


alainm23:
Thank you very much

PR #1326, opened 2024-06-03
Resolved Vala lint errors

JJeeff248:
Fixed errors in the code formatting picked up by the linter.

In `Util.vala` I ignored some issues raised with the regex as there is a [known issue](https://github.com/vala-lang/vala-lint/issues/141) and the workaround being ignoring it with `// vala-lint=space-before-paren`.

All other errors were resolved according to the error message.


alainm23:
Thank you very much

PR #1345, opened 2024-06-20
Update Dutch translation

Gert-dev:
Updates the Dutch translation to contain the new strings and some minor fixes to the existing ones here and there.


alainm23:
Thanks

Issue #1264, opened 2024-04-27
Compile instructions needs more details, unable to compile


meyyappan:
**Describe the bug**

I am trying to following the [compile instructions](https://github.com/alainm23/planify?tab=readme-ov-file#-compile) to compile planify from source. I installed the listed dependencies but when getting errors when running `meson` about missing dependencies.

**To Reproduce**

Working in Debian 12.

1. `sudo apt install libportal-gtk4-dev`
2. `sudo apt install libadwaita-1-dev`
3. `pip3 install meson`
4. `meson build --prefix/usr`

**Expected behavior**

Expect this step to succeed in compilation process.

**Actual behavior**

```sh
The Meson build system
Version: 1.4.0
Source dir: /home/user/planify
Build dir: /home/user/planify/build
Build type: native build
Project name: io.github.alainm23.planify
Project version: 4.6
C compiler for the host machine: cc (gcc 12.2.0 "cc (Debian 12.2.0-14) 12.2.0")
C linker for the host machine: cc ld.bfd 2.40

meson.build:1:0: ERROR: Could not execute Vala compiler: valac

A full log can be found at /home/user/planify/build/meson-logs/meson-log.txt
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
```

I install the dependencies `meson` complains is missing, but it would be easier if all the dependencies are listed above and instructions on how to install them. I install the dependencies it complains about manually but I'm stuck at `gxml-0.20`. Not sure how to install this.

```sh
Run-time dependency gxml-0.20 found: NO (tried pkgconfig and cmake)

meson.build:29:14: ERROR: Dependency "gxml-0.20" not found, tried pkgconfig and cmake
```

**Screenshots**

**Desktop (please complete the following information):**
 - OS or DE: debian 12

**Additional context**

Can you provide build instructions on building planify with all the necessary dependencies.

alainm23:
Hi and thank you for using Planify, you could build the project locally, but I recommend using Gnome Builder, just clone the project and Gnome Builder will build the project for you.

PR #1378, opened 2024-07-30
fix gxml wrap

peter-urban:
When trying to compile i was stuck at 
```
meson.build:29:14: ERROR: Dependency "gxml-0.20" not found, tried pkgconfig and cmake
```
This is the same error as described in https://github.com/alainm23/planify/issues/1264

I noticed you already added a gxml.wrap file in the subprojects directory, however for me meson did not pick it up (meson version 1.5.1, ubuntu 24.04) because the dependency searched for gxml-0.20

I made the following changes to the wrap file:

- renamed wrap to to gxml-0.20.wrap such that meson can find it with dependency('gxml-0.20')
- set revision to gxml-0.20 to make sure that this specific branch is checked out from the git repository
- add [provide] tag to export libxml_dep as dependency (otherwise there is a 'no gxml-0.20' dependency provided error)

On my computer it now successfully picks up the gxml-0.20 wrap in case libgxml-0.20 is not available on the machine

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,6 @@ build-aux/flatpak_build
 build-aux/.flatpak-builder
 build-aux/io.github.alainm23.planify.Devel.flatpak
 .flatpak-builder/
-subprojects/gxml
+subprojects/gxml**/
 *.snap
 planify*txt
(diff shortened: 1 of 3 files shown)


peter-urban:
update: I also renamed the subproject folder where the gxml git files are downloaded to gxml-0.20 to be more consistent and consequently updated the .gitignore to ignore all gxml**/ folders (not the wrap file) in the subprojects folder

alainm23:
Thank you very much

PR #1390, opened 2024-08-02
German translation finally done and tooltip error (at attachments)

McBaumwolle:
Sorry for the confusion, now the German translation should be completely done - nothing important, just some errors I missed here and there. 

The attachment list also had a delete button that had the tooltip "Download" - renamed that to "Delete" and tried to add a translation to `de.po` and also a margin because the box was a bit too much on the right. 

(I also renamed labels "Kennzeichnungen" to "Labels" because it is technically true in German, but nobody would ever use Kennzeichnungen.)

Unfortunately, I could not get Vala running on my end, so I also could not check if the margin is correct now.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -1979,13 +1979,13 @@ msgstr "Aktualisiert am"
 #: src/Dialogs/ItemChangeHistory.vala:62
 msgid ""
 "Your change history will be displayed here once you start making changes."
-msgstr ""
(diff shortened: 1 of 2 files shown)


alainm23:
Thank you very much, don't worry, I've fixed the bug, just a tip, when you change any translation of the source, you need to update the translations, for that use this command `ninja io.github.alainm23.planify-update-po`.

McBaumwolle:
Ohh thank you, I'll try that after I get the dependencies running.

PR #1392, opened 2024-08-02
po: Updated Hindi Translation

Scrambled777:
Minor updates to Hindi translation.


Scrambled777:
Hi @alainm23, is there any per-release announcements so we as a translator can get all strings updated in time for release?

Issue #1448, opened 2024-09-26
Deleted calendar events appear when adding new event


majjejjam:
**Describe the bug**
After adding an event via GNOME Calendar, it shows up in the scheduled view of Planify. When removing it from GNOME Calendar, it disappears. (So far so good.) When adding another event the same day (or month if further in the future), the event appears in the scheduled view, together with the deleted event.

**To Reproduce**
1. Add event in GNOME Calendar and watch it appear in the scheduled view
2. Remove the event from GNOME Calendar and watch it disappear
3. Add a new event the same day and watch it *and* the deleted event appear in the scheduled view

**Expected behavior**
The deleted event should not reappear when adding a new event.

**Screenshots**
![image](https://github.com/user-attachments/assets/68df0a71-dede-4670-bd99-a52bcceccc49)
Here, two deleted events previously added to November 5th are seen again.

**Desktop (please complete the following information):**
 - Fedora Silverblue with GNOME 47
 - Version 4.11.4

PR #1450, opened 2024-09-26
fix: captured deleted cal event

golamrabbiazad:
Fixed issue of https://github.com/alainm23/planify/issues/1448

If you have any feedback about this PR, please let me know.


alainm23 (commented):
Thank you for your PR, could it go inside the if?

golamrabbiazad:
Yes, `listbox` item should removed while `event_row` has value.

Thanks for the review. Much Appreciated :+1:

PR #1468, opened 2024-10-18
Update LINGUAS

Musikern:
Added Swedish language

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -10,3 +10,4 @@ nl
 hi
 bg
 ko
+sv


po/LINGUAS
```diff
@@ -10,3 +10,4 @@ nl
 hi
 bg
 ko
+sv
```
alainm23, line 13: If you want to add a new translation, you have to add the `sv.po` file as well.

alainm23 (changes requested):

Issue #1475, opened 2024-11-02
'Low priority' sorting / Normal priority default priority?


Gert-dev:
**Is your feature request related to a problem? Please describe.**
I usually add tasks with no priority since I usually rely on due dates and everything has to happen that day anyway. Recently however, I had a few tasks that have high as well as low priority as the latter are the best candidates for being postponed in case I find myself having too little time that day.

I see now that low priority items get sorted above items that have no priority at all, making it appear if tasks with low priority are more important than those with no priority. In a way this makes sense since I didn't mention the priority of the remaining tasks, but in my mind 'no priority' is equal to 'normal priority'.

**Describe the solution you'd like**
One of these could work:

* Normal priority items have no yellow colour assigned, since it's 'not special'.
* Sort 'No priority' on the same level as 'Normal priority'.
* Do away with 'No priority' and give all items normal priority by default, also removing the special colour for it (I think this is what Getting Things GNOME does).

**Describe alternatives you've considered**
* I could assign a normal priority to all tasks, which is a one-time undertaking, and then set the default priority in the settings to be 'Normal priority'. Unfortunately normal priority gets a colour assigned currently, which will make my task list a colouring book with yellow since this is what pretty much all my tasks will have - highlight colours should also be reserved for things that should stand out, making the priority colours less useful.
* Don't use priorities at all.
* Just live with it :wink: .

alainm23:
Hello @Gert-dev , and thank you for using Planify. I understand what you're saying, but it doesn't apply to priorities in Planify. There is no 'normal' priority, only low, medium, and high, and the way they are sorted depends on the hierarchy: high is always at the top, followed by medium and low. Tasks without a priority will always be at the bottom.

Gert-dev:
Thanks for the information. This made me realize why I am confused: I ended up translating 'medium' to what (in Dutch) essentially translates back to 'normal' in English. I should probably update the translation to avoid the confusion instead.

PR #1478, opened 2024-11-08
Update Dutch translation

Gert-dev:
Updates the Dutch translation to contain the latest new strings.

Also fixes the confusion I mentioned in https://github.com/alainm23/planify/issues/1475#issuecomment-2464808820.


alainm23:
Thank you very much

Issue #1489, opened 2024-11-26
Plainify Freezes When Google Pings Blocked


amd1890:
**Describe the bug**
Planify from the latest flapak repo sometimes just stops responding. I am also using Pop!_OS 22.04 LTS. I also started blocking Planify from accessing the internet. I may have been because planify seemed to be pinging google, but I could be wrong, my memory is bad sometimes. Systemd was pinging google at a certain point too so this could be incorrect, but I thought I saw planify request to check google.com. I have also listed hyperlinks in the program and am not sure if this could be why. Even if it is incorrect, the freezes are real. I have blocked Planify from accessing the Internet and it does freeze. With this block, Planify seems to regularly stop working, the window won't work, I can move it, and I get a pop up asking me to wait or close the App.

**To Reproduce**
1. Download Planify from flatpak
2. Download Opensnitch
3. Block Planify from Accessing Internet
4. Planify glitches

**Expected behavior**
Planify running despite internet block

**Screenshots**
I don't have them

**Desktop (please complete the following information):**
 - OS or DE: gnome
 - Version: 42.9

**Additional context**
I really like Planify a lot. It's very visual, intuitive, and makes breaking down tasks into categories and subtasks with priorities simple. I wish I knew why it was pinging google or if I am not understanding what is going on. I am not a developer and so even with open source code I am often trusting other developers to review code or that the code is good. It's also possible Planify does everything it says it does and I don't understand the situation. I have limited skills and intelligence.

alainm23:
Hi, Planify pings google to verify the internet access, anyway I will check why is the bug you mentioned happening.

amd1890:
Thank you! I like planify a lot! In a future version, would it be possible to allow turning off pings in settings or letting the user ping something not associated with big tech? I don't dislike big tech but don't trust their 1 jillion trusted partners. I created a cronjob to run planify @reboot and I'd prefer google didn't get my ping time with every reboot, even if it may not log that and feed it to an AI to try to figure out things about me, it probably doesn't do that although who knows. Thanks for making Planify!

byquanton:
@amd1890 this issue is fixed since version [4.12.0](https://github.com/alainm23/planify/releases/tag/4.12.0) or specific the pr #1511

PR #1511, opened 2025-01-14
Rewrite Offline Mode 

byquanton:
This PR removes the current online detection and replaces it with a callback that shows the user when a connection to a server couldn't be made. 
Would fix the following issue: https://github.com/alainm23/planify/issues/1489

Things left to-do in the current implementation:
- [x] Add sync_failed calls to Todoist
- [x] Update translations/Finalize the new text
- [ ] ~~Make it possible to add to-dos offline~~ -> Create a new PR for this

If you have any feedback about this PR, please let me know.


byquanton:
New To-Do: Make it possible to add a To-Do while the server is offline. Maybe add an icon behind tasks when changes aren't synchronized yet
image

src/Widgets/SyncButton.vala
byquanton, line 1: Ideally this should be a single button with a changing icon_name, but I haven't managed to get it to work that way without breaking the transition.

alainm23: Currently the offline insertion works with Todoist, it needs to be developed for Nextcloud.But the ideal would be to work it in another PR, so as not to make the revision too big.

Fadikkop:
The Notifications of the Flatpak are per default denied.

alainm23 (approved):

alainm23:
Thank you very much

PR #1580, opened 2025-06-02
Document architecture

EthanDawes:
Added some documentation to help newcomers. Willing to make revisions as necessary.

Co-authored by @kimjammer @Xendergo @Cate701


alainm23:
Oh, good job, I don't think even I could have done it better. Honestly, I haven’t had much time to work on Planify lately. I stopped using Linux a while ago because of work.

PR #1603, opened 2025-07-13
Use Nextcloud Login Flow for login

### LINKED ISSUE #1595: Nextcloud: Support Login Flow V2
**Is your feature request related to a problem? Please describe.**
Some Nextcloud instances utilize external authentication providers, which means users do not have direct login credentials to enter into Planify.

**Describe the solution you'd like**
Implement Login Flow V2, which leverages the user's default system browser for authentication and supports external authentication within Nextcloud.

**Describe alternatives you've considered**
Currently, users must create an App Password in their Nextcloud instance, which is less intuitive.

**Additional context**
Docs: https://docs.nextcloud.com/server/latest/developer_manual/client_apis/LoginFlow/index.html#login-flow-v2
 If I find some time, I will attempt to implement this feature myself.

byquanton:
Fixes #1595

### DIFF
--- a/core/Constants.vala
+++ b/core/Constants.vala
@@ -20,6 +20,7 @@
 */
 
 namespace Constants {
+    public const string SOUP_USER_AGENT = "Planify";
     public const string TODOIST_CLIENT_ID = "b0dd7d3714314b1dbbdab9ee03b6b432";
     public const string TODOIST_CLIENT_SECRET = "a86dfeb12139459da3e5e2a8c197c678";
     public const string TODOIST_SCOPE = "data:read_write,data:delete,project:delete";
--- a/core/Services/CalDAV/Providers/Nextcloud.vala
+++ b/core/Services/CalDAV/Providers/Nextcloud.vala
@@ -370,8 +370,7 @@ public class Services.CalDAV.Providers.Nextcloud : Services.CalDAV.Providers.Bas
 					break;
 				}
 
-                // TODO: Rewrite this function with Timeout.add_seconds ?
-				yield nap (interval * 1000);
+				yield Util.nap (interval * 1000);
 
 				timeout -= interval;
 			}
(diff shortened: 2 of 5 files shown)


byquanton:
https://github.com/user-attachments/assets/d8de7b4b-b97b-4b47-a9d9-3d5a4d38cafb

core/Services/CalDAV/Providers/Nextcloud.vala
```diff
@@ -293,6 +301,101 @@ public class Services.CalDAV.Providers.Nextcloud : Services.CalDAV.Providers.Bas
...
+					}
+					
+				} catch (Error err) {
+					response.error_code = err.code;
+					response.error = "Polling error: %s".printf (err.message);
+					break;
+				}
+
+                // TODO: Rewrite this function with Timeout.add_seconds ?
+				yield nap (interval * 1000);
```
byquanton: @alainm23 The only thing I'm unsure with this implementation is if the while loop should be replaced with a recursive function that uses Glib.Timeout or is this fine? It isn't blocking the main thread because its async.

core/Constants.vala
```diff
@@ -20,6 +20,7 @@
 */
 
 namespace Constants {
+    public const string SOUP_USER_AGENT = "Planify";
```
byquanton, line 23: Should the User Agent be added to other Soup Messages as well?

alainm23:
Thank you for this development. Unfortunately, I don't currently have a Linux machine available to test it, but if you believe everything is correct, we can go ahead and merge it without any problem.

core/Services/CalDAV/Providers/Nextcloud.vala
alainm23: Ideally, it would be best to use a Glib.Timeout to avoid blocking the app's UI.

core/Constants.vala
alainm23: We could leave it as an improvement for later — I think it's not necessary for now.

core/Services/CalDAV/Providers/Nextcloud.vala
byquanton: It isn't blocking the app's UI. The "nap" method is implemented below and is just an async wrapper of Glib.Timeout. I tough about rewriting the method to be using the Timeout directly instead of using a while loop and sleeping in the thread.

byquanton: The nap method is adapted from [https://wiki.gnome.org/Projects/Vala/AsyncSamples (Async sleep example)](https://wiki.gnome.org/Projects/Vala/AsyncSamples#Async_sleep_example).

byquanton:
I moved the nap method into the 'Util' method so that it can be used by other methods in the future.

I have been using Planfiy with this branch for the last few days and have not encountered any other issues. Therefore, from my perspective, it can be merged.

PR #1606, opened 2025-07-14
Fix get_short_name method to correctly handle variable length UTF-8 strings

### LINKED ISSUE #1583: Deleting task with specific UTF-8 string name cause the app to crash.
**Describe the bug**
Deleting tasks with some specific UTF-8 string name causes the app to crash (killed by SIGSEGV). 

Examine the log, the error is created at the banner for deletion, with log:
```
(io.github.alainm23.planify:2): Gtk-WARNING **: 21:05:24.142: Failed to set text '阶段性成果：\xe8\x8b… was deleted' from markup due to error parsing markup: Error at line 1, character 44: name contains invalid UTF-8 encoded text - invalid '阶段性成果：\xe8\x8b… was deleted'
```

The task name is however indeed a valid UTF-8 string. 

Upon examination, The problem lies at 
https://github.com/alainm23/planify/blob/42319d05d174a0920e5584244d81bc25979735e0/src/Layouts/ItemBoard.vala#L1042 

and deeper at
https://github.com/alainm23/planify/blob/42319d05d174a0920e5584244d81bc25979735e0/core/Util/Util.vala#L367-L375

This slice truncates on the byte level, not per character, and slices a valid Unicode character `英` (UTF-8 is a variable length coding, thus a 3-byte codepoint `E8 8B B1`) unable to decode. 

**To Reproduce**
1. Create a task with name `阶段性任务：英语`；
2. Try to delete it.
3. Crash triggered.

**Expected behavior**
The crash is not desired. The trimming is clearly against the usage of UTF-8 encoding; Furthermore, such minute GUI rendering error should at most cause a mojibake or be caught by other exceptions.

**Screenshots**
NA

 - OS or DE: Fedora 42, GNOME 48 on Wayland.
 - Version: 4.12.0

**Additional context**
NA

byquanton:
Fixes https://github.com/alainm23/planify/issues/1583

### DIFF
--- a/core/Util/Util.vala
+++ b/core/Util/Util.vala
@@ -367,8 +367,9 @@ public class Util : GLib.Object {
     public string get_short_name (string name, int size = Constants.SHORT_NAME_SIZE) {
         string returned = name;
         
-        if (name.length > size) {
-            returned = name.slice (0, size) + "…";
+        int char_count = name.char_count ();
+        if (char_count > size) {
+            returned = name.substring (0, name.index_of_nth_char (size)) + "…";
         }
 
         return returned;


byquanton:
It now uses 'char_count' to compare sizes and 'index_of_nth_char' to get the correct index for creating a substring.

alainm23 (approved):

PR #1674, opened 2025-08-12
Show project title in HeaderBar on scroll

### LINKED ISSUE #1589: Non-ideal icon usage for toggling the left sidebar
Issue: The "closed" state icon for the left sidebar uses the icon for the right sidebar which does not reflect the position of the left sidebar. The action toggles the left sidebar not right one. It's a bit confusing since the app features a right sidebar but the button toggles the left sidebar instead.

https://github.com/user-attachments/assets/4b4e7270-9791-4f67-9751-4e4994f11aac

Solution: Use the same icon for each state for the left sidebar. This is the behavior of multiple GNOME apps which have a sidebar such as Cartridges and Builder (and probably more).

https://github.com/user-attachments/assets/5eef7566-221d-49c8-a396-48def7acdf1e

alainm23:
When viewing a project, the title is now shown in the HeaderBar once the main title in the content area is scrolled out of view. This improves navigation and maintains context while scrolling.

Fixes: #1589

https://github.com/user-attachments/assets/23c7044c-5d79-4577-bb83-14cd29635293

### DIFF
--- a/core/Constants.vala
+++ b/core/Constants.vala
@@ -45,4 +45,5 @@ namespace Constants {
     public const bool SHOW_WHATSNEW = false;
     public const bool BLOCK_PAST_DAYS = false;
     public const int COMPLETED_PAGE_SIZE = 15;
+    public const int HEADERBAR_TITLE_SCROLL_THRESHOLD = 24;
 }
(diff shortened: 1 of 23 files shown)


ztefn:
That's a very nice solution! I'm gonna have to steal this concept for the titles in Haguichi's sidebar... 🤫

PR #1677, opened 2025-08-14
Refine Sources view and remove redundant popover button

alainm23:
This PR updates the Sources view with visual adjustments:

- Removed the popover button (which only contained the delete option) and integrated the delete action directly into the source view.
- Adjusted the layout so each source type only displays its relevant elements.

These changes simplify the interface and improve clarity in the Sources view.

image

image


ztefn:
Some suggestions to streamline the sources list further:

- Add drag handle image (`list-drag-handle-symbolic` with class `dimmed`) on the left side to indicate the rows are draggable (then you can drop the text _You can sort your accounts by dragging and dropping_)
- Use switch instead of checkbox to turn sources on/off
- Make all rows the same height (the ones without subtitle are now smaller than with subtitle in you screenshot)

Example list featuring drag handles and switches:
preferences-commands-2x

alainm23:
Thank you, I’ll keep that in mind

PR #1683, opened 2025-08-15
Retain the last used project in Quick Add

xxfogs:
As the title suggests.

Such mechanism already exists for _Quick Add_ but not for the same dialog within the app. I have found that feature to be especially convenient.


alainm23:
Makes sense to me, thank you so much.

PR #1739, opened 2025-08-31
Check if key is valid color

byquanton:


byquanton:
i broke my cherry pick

byquanton:
now its correct

alainm23 (approved):

PR #1747, opened 2025-09-02
Add support for detecting deleted task lists in CalDAV and Nextcloud

alainm23:
Implements automatic detection of deleted task lists to keep local database synchronized with CalDAV servers.

Changes

- Generic CalDAV: Detects deletion by comparing server URLs with local projects
- Nextcloud: Added support for `deleted-calendar` resourcetype
- Auto cleanup: Automatically removes deleted projects from local database

Implementation
- Updated PROPFIND XML to include Nextcloud namespace
- Added `is_deleted_calendar()` function for detection
- Enhanced sync logic to handle both generic CalDAV and Nextcloud deletion patterns

Fixes sync issues when task lists are deleted from other clients or web interface.

### DIFF
--- a/core/Objects/Project.vala
+++ b/core/Objects/Project.vala
@@ -254,11 +254,9 @@ public class Objects.Project : Objects.BaseObject {
         }
     }
 
-    // The old implementation checked for "deck--board" in the id, this no longer works.
-    // TODO: Find found why there was seperation for deck projects, maybe find another way to detect it.
     public bool is_deck {
         get {
-            return false;
+            return "deck--board" in calendar_url.down ();
         }
     }
 
--- a/core/Services/CalDAV/CalDAVClient.vala
+++ b/core/Services/CalDAV/CalDAVClient.vala
@@ -165,14 +165,14 @@ public class Services.CalDAV.CalDAVClient : Services.CalDAV.WebDAVClient {
 
     public async void sync (Objects.Source source, GLib.Cancellable cancellable) throws GLib.Error {
         var xml = """<?xml version='1.0' encoding='utf-8'?>
-                    <d:propfind xmlns:d="DAV:" xmlns:ical="http://apple.com/ns/ical/" xmlns:cal="urn:ietf:params:xml:ns:caldav">
+                    <d:propfind xmlns:d="DAV:" xmlns:ical="http://apple.com/ns/ical/" xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:nc="http://nextcloud.com/ns">
                         <d:prop>
                             <d:resourcetype />
                             <d:displayname />
                             <d:sync-token />
                             <ical:calendar-color />
                             <cal:supported-calendar-component-set />
-                            <x3:deleted-at xmlns:x3="http://nextcloud.com/ns"/>
+                            <nc:deleted-at/>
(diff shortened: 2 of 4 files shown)


core/Objects/Project.vala
```diff
@@ -254,11 +254,9 @@ public class Objects.Project : Objects.BaseObject {
         }
     }
 
-    // The old implementation checked for "deck--board" in the id, this no longer works.
-    // TODO: Find found why there was seperation for deck projects, maybe find another way to detect it.
     public bool is_deck {
         get {
-            return false;
+            return "deck--board" in calendar_url.down ();
```
byquanton, line 259: What is the is_deck check used for? Why does Planify need to limit the functionality?

core/Services/CalDAV/CalDAVClient.vala
```diff
@@ -172,23 +172,49 @@ public class Services.CalDAV.CalDAVClient : Services.CalDAV.WebDAVClient {
                             <d:sync-token />
                             <ical:calendar-color />
                             <cal:supported-calendar-component-set />
+                            <x3:deleted-at xmlns:x3="http://nextcloud.com/ns"/>
```
byquanton: I would add the namespace at the top, and also call the namespace "nc" instead of "x3"

core/Services/CalDAV/CalDAVClient.vala
```diff
@@ -172,23 +172,49 @@ public class Services.CalDAV.CalDAVClient : Services.CalDAV.WebDAVClient {
...
-        // TODO: Implement check for deleted calendars
+
+        // Delete CalDAV Generic
+        var server_urls = new Gee.HashSet<string> ();
+        foreach (var response in multi_status.responses ()) {
+            if (response.href != null) {
+                server_urls.add (get_absolute_url (response.href));
+            }
+        }
+
```
byquanton, line 190: Looks good, I will check if this works.

core/Objects/Project.vala
byquanton: Otherwise, this looks like a good solution. I haven't found anything in the current propfind response that would be a better hint.

core/Services/CalDAV/CalDAVClient.vala
byquanton: Change `<d:propfind xmlns:d="DAV:" xmlns:ical="http://apple.com/ns/ical/" xmlns:cal="urn:ietf:params:xml:ns:caldav">` into
`<d:propfind xmlns:d="DAV:" xmlns:ical="http://apple.com/ns/ical/" xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:nc="http://nextcloud.com/ns">`

and then just use
`<nc:deleted-at/>`

byquanton (changes requested):
The only thing I would definitely change is the Nextcloud namespace naming. Otherwise, it looks good.

core/Objects/Project.vala
alainm23: When a task list is of type `deck`, it’s not possible to create, edit, or delete tasks—only read them. In this sense, Planify is able to disable several functions in the project view.

byquanton: Okay, makes sense

PR #1754, opened 2025-09-06
Make Webkit Optional

byquanton:


alainm23 (approved):

alainm23:
Great work, I was planning to do something like that later.

PR #1780, opened 2025-09-15
Fix CalDAV connection issue with Baïkal by adding an option to bypass resolve_well_known_caldav()

nijakow:
When connecting to a custom Baïkal instance (v0.10.1, running via HTTP on port 8543), planify fails due to an incorrect `dav_endpoint` being computed.

Baïkal returns a `Location` header that omits the correct port, so `resolve_well_known_caldav()` produces a wrong endpoint (e.g. `http://<host>/...` instead of `http://<host>:8543/...`), causing socket timeouts and "unknown errors" in `do_login()`.

This PR adds an option to bypass `resolve_well_known_caldav()` as a temporary workaround, so that planify can successfully connect to Baïkal instances via HTTP using non-standard ports.

### DIFF
--- a/src/Dialogs/Preferences/Pages/Accounts/CalDAVSetup.vala
+++ b/src/Dialogs/Preferences/Pages/Accounts/CalDAVSetup.vala
@@ -32,6 +32,7 @@ public class Dialogs.Preferences.Pages.CalDAVSetup : Dialogs.Preferences.Pages.B
     // Advanced Options
     private Adw.EntryRow calendar_home_entry;
     private Widgets.IgnoreSSLSwitchRow ignore_ssl_row;
+    private Widgets.BypassResolveSwitchRow bypass_resolve_row;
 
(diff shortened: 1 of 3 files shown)


alainm23:
The best option UX-wise would be to make this automatic: if `resolve_well_known_caldav()` fails or returns an endpoint without the original port, Planify should reconstruct the endpoint preserving the host:port. That way misconfigured servers work out of the box, without forcing users to toggle advanced settings

byquanton (approved):

byquanton:
> The best option UX-wise would be to make this automatic: if `resolve_well_known_caldav()` fails or returns an endpoint without the original port, Planify should reconstruct the endpoint preserving the host:port. That way misconfigured servers work out of the box, without forcing users to toggle advanced settings

I know there are some providers that do redirects to other domains (and ports), so this solution provides a fix for misconfigured Baikal instances (An issue that originates from Baikal-Docker).

PR #1784, opened 2025-09-15
Make Planify compilable on FreeBSD and Windows.

byquanton:
This makes it possible to compile Planify on: FreeBSD, Windows & *macOS*.

macOS is crashing currently, could also be related to missing packages I didn't install with homebrew

### DIFF
--- a/src/Dialogs/Preferences/Pages/General.vala
+++ b/src/Dialogs/Preferences/Pages/General.vala
@@ -73,6 +73,7 @@ public class Dialogs.Preferences.Pages.General : Dialogs.Preferences.Pages.BaseP
 
         // de_group.add (run_background_row);
 
+        #if WITH_LIBPORTAL
         var run_on_startup_switch = new Gtk.Switch () {
             valign = Gtk.Align.CENTER,
             active = Services.Settings.get_default ().settings.get_boolean ("run-on-startup")
@@ -85,6 +86,7 @@ public class Dialogs.Preferences.Pages.General : Dialogs.Preferences.Pages.BaseP
         run_on_startup_row.add_suffix (run_on_startup_switch);
 
         de_group.add (run_on_startup_row);
+        #endif
 
         var calendar_events_switch = new Gtk.Switch () {
             valign = Gtk.Align.CENTER,
@@ -170,9 +172,11 @@ public class Dialogs.Preferences.Pages.General : Dialogs.Preferences.Pages.BaseP
             Services.Settings.get_default ().settings.set_boolean ("run-in-background", run_background_switch.active);
         })] = run_background_switch;
 
+        #if WITH_LIBPORTAL
         signal_map[run_on_startup_switch.notify["active"].connect (() => {
             Services.Settings.get_default ().settings.set_boolean ("run-on-startup", run_on_startup_switch.active);
         })] = run_on_startup_switch;
+        #endif
 
         signal_map[calendar_events_switch.notify["active"].connect (() => {
             Services.Settings.get_default ().settings.set_boolean ("calendar-enabled", calendar_events_switch.active);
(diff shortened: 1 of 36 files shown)


byquanton:
This PR only contains minor patches. These can be merged already. The current state is that Planify compiles on FreeBSD, Windows, and macOS. There are still many issues with Windows and even more with macOS, as it currently crashes with the default project on my system.

Planify works without issues on FreeBSD when compiled without Evolution.

@alainm23 can you review if my commit ffa671e28a306cdca99dcb37a80d7d55c93f2dbf is right? Because I don't see a reason for the method to still exists.

alainm23:
Excellent work, it looks really good to me

> @alainm23 can you review if my commit https://github.com/alainm23/planify/commit/ffa671e28a306cdca99dcb37a80d7d55c93f2dbf is right? Because I don't see a reason for the method to still exists.

If it’s no longer necessary, I plan to refactor that part of the filters later

src/Dialogs/Preferences/Pages/General.vala
```diff
@@ -170,9 +172,11 @@ public class Dialogs.Preferences.Pages.General : Dialogs.Preferences.Pages.BaseP
             Services.Settings.get_default ().settings.set_boolean ("run-in-background", run_background_switch.active);
         })] = run_background_switch;
 
+#if WITH_LIBPORTAL
         signal_map[run_on_startup_switch.notify["active"].connect (() => {
             Services.Settings.get_default ().settings.set_boolean ("run-on-startup", run_on_startup_switch.active);
         })] = run_on_startup_switch;
+// #endif (some how vala has issues when endif is here?!?!)
```
alainm23: This is throwing an error at compile time

alainm23:
I ran some tests on a macOS machine with an M1 processor and got several errors, especially with `libical`, since the `.vapi` file was missing. I understand this is still a work in progress. I’ll finish testing on a Linux machine to make sure nothing is broken, and then we can work on fixes specific to each platform.

alainm23 (changes requested):

byquanton:
> I ran some tests on a macOS machine with an M1 processor and got several errors, especially with `libical`, since the `.vapi` file was missing. I understand this is still a work in progress. I’ll finish testing on a Linux machine to make sure nothing is broken, and then we can work on fixes specific to each platform.

You have to build a custom libical with the vapi files. Here is what I changed in the brew file locally:

```ruby
[...]
  def install
    args = %W[
      -DBDB_LIBRARY=BDB_LIBRARY-NOTFOUND
      -DENABLE_GTK_DOC=OFF
      -DSHARED_ONLY=ON
      -DBUILD_SHARED_LIBS=ON
      -DGOBJECT_INTROSPECTION=ON
      -DICAL_GLIB_VAPI=ON
      -DICAL_GLIB=ON
      -DCMAKE_INSTALL_RPATH=#{rpath}
    ]

    system "cmake", "-S", ".", "-B", "build", *args, *std_cmake_args
    system "cmake", "--build", "build"
    system "cmake", "--install", "build"
  end
[...]
 ```

src/Dialogs/Preferences/Pages/General.vala
byquanton: Okay, adding it back does throw an error too ...

i don't get it

byquanton: So, when building without libportal, the #endif is needed, but when building with it, the #endif somehow breaks things.

byquanton: The Vala preprocesso chokes with this code syntax, I've split this up and now the preprocessor is happy.

```vala
        signal_map[run_on_startup_switch.notify["active"].connect (() => {
            Services.Settings.get_default ().settings.set_boolean ("run-on-startup", run_on_startup_switch.active);
        })] = run_on_startup_switch;
 ```

PR #1886, opened 2025-10-08
Fix for sidebar reordering crash

### LINKED ISSUE #1805: Sorting side panel in setting causing crash
**Describe the bug**
Drag&Drop action in App Setting -> Side Panel -> Display or Hide Side Panel section would cause the app to crash silently. 
Additionally, closing the App Setting dialog when on this page would also cause the app to crash.
No terminal log emitted.

**To Reproduce**

https://github.com/user-attachments/assets/720948e8-3efc-4c9b-8338-2c8787c4e2db

**Expected behavior**
App should not crash.

**Screenshots**

See To Reproduce section.

 - OS or DE: Fedora 42 Workstation, GNOME DE.
 - Version: 4.14.0

**Additional context**
NA

CodedOre:
This attempts to provide a fix for issue #1805.

This changes the following:

- In the `clean_up` method of `SidebarRow`, the `select_gesture` member was removed before it was supposed to be used to disconnect signals, causing a SEGFAULT. The fix changes the order so that the signal disconnect happens before `select_gesture` is removed.
- In the `on_drop` method of `ReorderChild`, the row seems to be cleaned up by GObject between it being removed and re-added to the list, causing a SEGFAULT. The fix adds an explicit reference, which seems to keep the row in memory, preventing the SEGFAULT.

Fixes #1805

### DIFF
--- a/src/Dialogs/Preferences/Pages/Sidebar.vala
+++ b/src/Dialogs/Preferences/Pages/Sidebar.vala
@@ -317,16 +317,15 @@ public class Dialogs.Preferences.Pages.Sidebar : Dialogs.Preferences.Pages.BaseP
                 reorder = null;
             }
 
+            foreach (var entry in signal_map.entries) {
+                entry.value.disconnect (entry.key);
+            }
+            signal_map.clear ();
+
             if (select_gesture != null && action_row != null) {
                 action_row.remove_controller (select_gesture);
                 select_gesture = null;
             }
-
-            foreach (var entry in signal_map.entries) {
-                entry.value.disconnect (entry.key);
-            }
-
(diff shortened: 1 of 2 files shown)


CodedOre:
These two fixes seem to resolve the issue of Planify crashing when reordering the sidebar in the settings. However, while reordering works, there is currently an warning printed:

```
(io.github.alainm23.planify.Devel:2): Gtk-WARNING **: 15:18:14.086: (../gtk/gtkdroptargetasync.c:271):gtk_drop_target_async_handle_crossing: runtime check failed: (self->drop == NULL || self->drop == crossing->drop)
```

Issue #1427, opened 2024-09-04
Tag adding while adding a task w/ the @ operator 


IslamTayeb:
In normal todoist, you're able to put information about the priority, tags, and dates when writing the title of a task (e.g. Test p1 @Personal would make a task that has the title as "Test" with Priority 1 and a tag of "Personal"). It'd be beneficial to have this as it makes the process of task creation a lot more convenient.

aadrian:
@IslamTayeb 
Maybe using the '#' instead of '@'  ? ( since '#' hashtag is used on most platforms for tagging ? )

Heinrich-XIAO:
> [@IslamTayeb](https://github.com/IslamTayeb) Maybe using the '#' instead of '@' ? ( since '#' hashtag is used on most platforms for tagging ? )

Todoist uses @ for labels.

PR #1971, opened 2025-10-28
Add @ symbol label picker with keyboard navigation and auto-scroll

### LINKED ISSUE #1881: Improve keyboard-based label selection when adding a task
I love that typing `@` brings up the label list when creating a task - that's a very intuitive feature. However, I feel that the process of selecting a label could be improved. Currently, I often have to use the mouse to select a label, which disrupts a keyboard-focused workflow. For example, this happens when adding multiple tasks one after the other or when using the quick-add command from outside the main UI. 

Here's what currently already works: 
- Type `@` to open the label drop-down menu
- The focus is then set to the search text box, allowing to filter existing labels (and create new ones, if there are no matches)
- Pressing the `Tab` key shifts the focus to the filtered label list and pressing 'Enter' selects/deselects the marked label
- Pressing `Esc` closes the label selection menu
 
Here’s what I’d expect and what would feel more intuitive to me:
1. After opening the selection box, I should be able to use the up and down arrows to navigate the label list. I feel this is more intuitive than using the `Tab` key.
2. `Space` could be used to select/deselect a label from the list. This would free up `Enter` for `Select label and Close the selector` functionality.
3. After typing something into the filter text field, the focus should automatically shift to the selection when the result is narrowed down to exactly one item (currently similar to hitting `Tab`). While I think this would be intuitive, I can see how this might result in unwanted behavior (e.g. the backspace key would no longer be available for deleting the filter text, or creating new labels is getting more difficult etc.).
4. Pressing `Esc` should close the drop-down menu and return focus to the task title field.

Original issue: #1813

alainm23:
Implements inline label selection using @ symbol trigger with:

- Real-time filtering as you type after @
- Keyboard navigation (arrow keys + Enter)
- Auto-scroll to follow selection in long lists
- Click outside to close popover
- Proper cursor positioning after label selection
- Clickable labels for removal

Enhances task creation workflow with quick label assignment similar to modern text editors.

Fixes: #1881 #1427

https://github.com/user-attachments/assets/0457a2e1-9a1a-4bf1-9145-66ca04c0d8a1


fabianschoeppach:
Wow, this looks so awesome. Thank you so much!

Issue #1958, opened 2025-10-23
Can't Rearrange Labels


MrDaisyBates:
**Describe the bug**
When on the "Labels" tab (CTRL+L), it seems I can grab them with my mouse, but it doesn't seem I have the ability to actually rearrange them.

**To Reproduce**
1. CTRL+L
2. Left Mouse Click a Label (and hold)
3. Drag the Label to a different spot on the list

**Expected behavior**
To be able to change the ordering of labels via click and drag.

**Desktop (please complete the following information):**
 - Fedora 42, Nvidia, Flatpak

PR #2001, opened 2025-11-05
Lables can be rearranged by click and drag

arsca96:
I noticed the same problem that was mentioned in [this](https://github.com/alainm23/planify/issues/1958) issue. The widget that the controller was being connected to was never receiving the drop, but the parent widget was. Now the controller is connected to the parent widget and it works.

### DIFF
--- a/src/Widgets/ReorderChild.vala
+++ b/src/Widgets/ReorderChild.vala
@@ -201,7 +201,7 @@ public class Widgets.ReorderChild : Adw.Bin {
             drag_source = null;
         }
 
-        if (drop_order_target != null && widget != null){
+        if (drop_order_target != null && widget != null) {
             widget.remove_controller (drop_order_target);
             drop_order_target = null;
         }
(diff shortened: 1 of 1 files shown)


src/Widgets/ReorderChild.vala
```diff
@@ -206,14 +201,9 @@ public class Widgets.ReorderChild : Adw.Bin {
...
 
-        if (drop_order_top_target != null && motion_top_grid != null) {
-            motion_top_grid.remove_controller (drop_order_top_target);
-            drop_order_top_target = null;
-        }
-
-        if (drop_order_bottom_target != null && motion_bottom_grid != null) {
-            motion_bottom_grid.remove_controller (drop_order_bottom_target);
-            drop_order_bottom_target = null;
+        if (drop_order_target != null && widget != null){
```
alainm23:
```suggestion
        if (drop_order_target != null && widget != null) {
```

alainm23 (changes requested):

alainm23 (approved):

PR #2049, opened 2025-11-16
Changed Portuguese (Portugal) .po to pt_PT

rottenpants466:
For consistency and for Planify to actually use the Portuguese (Portugal) language in a system using that language, we should rename the pt.po file to pt_PT.po, just like we do for the Portuguese (Brazil) translation pt_BR.po.

This change should also be made in weblate because, I think, it uses the name pt.po.

Thank you


alainm23:
Thanks for the PR!

I get what you mean about keeping things consistent with pt_BR, but in this case pt.po is actually the correct name. In gettext/Meson/Weblate setups, pt already stands for Portuguese from Portugal. Changing it to pt_PT.po would break language detection and also cause issues in Weblate, since it expects the file to stay as pt.po.

So for now it’s better to keep the current naming to avoid problems.

rottenpants466:
> Thanks for the PR!
> 
> I get what you mean about keeping things consistent with pt_BR, but in this case pt.po is actually the correct name. In gettext/Meson/Weblate setups, pt already stands for Portuguese from Portugal. Changing it to pt_PT.po would break language detection and also cause issues in Weblate, since it expects the file to stay as pt.po.
> 
> So for now it’s better to keep the current naming to avoid problems.

No problem!

I made that sugestion because of an issue i had with another app, Zapzap.

I added the translation, it got merged and the app got updated with the new translation. The issue was that the zapzap app was still using english instead of Portuguese Portugal: https://github.com/rafatosta/zapzap/issues/466

The mantainer changed the .po file to pt_PT and that fixed the issue.

The same thing happens with Planify, it's in english and not in my native language :|

Also, I completed the Planify pt translation on Weblate. Does the translation appear here automatically or do I have to do something there?

Thank you!

alainm23 (approved):

PR #2058, opened 2025-11-20
Feat/macOS install - EXPERIMENTAL

ngtech:
Long-time Linux user here who switched to macOS only a few days ago. 

Planify is one of the apps I rely on the most, so getting it running on macOS became my first priority. 

With some help from ChatGPT / codex for navigating macOS quirks, I managed to  get it to compile successfully and produce a working .dmg

I will continue working on this as free time allows.


alainm23:
Oh, thanks, this looks really good. I also use macOS for work and have been making a few adjustments so Planify can run on macOS. Honestly, I know very little about the platform. Another detail I noticed is that GLib.Signals aren’t working or are failing to fire, which is why many of Planify’s features don’t work — it’s strange, and I still need to investigate it thoroughly, which I haven’t done yet. But having a script to generate the .dmg is a big step forward and would help a lot with users who aren’t very experienced with the terminal

jzaehrin:
Currently on 26.1, I achived to compile the project. But planify does not start with the error `No such backend: quartz` from Gtk. Any idea ?

alainm23:
There are still some small details that need to be resolved. This PR lays the foundation, but it’s not fully functional yet. I hope I can find the time to address these issues and provide full support for macOS

PR #2159, opened 2025-12-30
Generate translation metrics based on installed translations

byquanton:
This PR introduces a new system for calculating translation metrics, replacing the previous approach that relied on fetching a JSON file from the repository that was generated in a Python script using GitHub Actions.


byquanton:
Instead of rebasing every time the bot updates the ‎data/translations_metrics.json file I decided to remove the deletion from this PR.

alainm23 (approved):

alainm23:
Thanks for your time, and sorry I couldn’t review this earlier — I was traveling and had limited internet access.

Regarding the PR, it works well and in the same way it did before; the difference is that it now works offline.

PR #2180, opened 2026-01-09
Don't update count during project sync

byquanton:


core/Services/Todoist.vala
```diff
@@ -357,6 +361,12 @@ public class Services.Todoist : GLib.Object {
         }
 
         source.last_sync = new GLib.DateTime.now_local ().to_string ();
+        foreach (var project in Services.Store.instance ().get_projects_by_source (source.id)) {
```
alainm23, line 364: We could reuse the same project list used on line `287` and avoid calling get_projects_by_source a second time

alainm23 (changes requested):

alainm23:
Everything works as expected — just one small comment and we can proceed

PR #2183, opened 2026-01-09
Improve Database operations for import and deleting projects

byquanton:
The commit from #2180 is already included in this PR 

- **Batch insert items when importing a project**
- **Batch delete items when deleting the project**

### DIFF
--- a/core/Objects/Project.vala
+++ b/core/Objects/Project.vala
@@ -531,6 +531,23 @@ public class Objects.Project : Objects.BaseObject {
         }
     }
 
+    public void add_items_batched (Gee.ArrayList<Objects.Item> items) {
+        foreach (var item in items) {
+            string? parent_id = Util.find_string_value ("RELATED-TO", item.calendar_data);
+            if (parent_id != null && parent_id != "") {
+                Objects.Item ? parent_item = Services.Store.instance ().get_item (parent_id);
+                if (parent_item != null) {
+                    parent_item.add_item_if_not_exists (item);
+                    items.remove (item);
+                }
+            }
+        }
+
+        if (!Services.Store.instance ().insert_items_transaction (items)) {
+            error ("Failed to insert items in transaction");
+        }
+    }
+
     public Objects.Item add_item_if_not_exists (Objects.Item new_item, bool insert = true) {
         Objects.Item ? return_value = null;
         lock (_items) {
--- a/core/Objects/Project.vala
+++ b/core/Objects/Project.vala
@@ -532,15 +532,21 @@ public class Objects.Project : Objects.BaseObject {
     }
 
     public void add_items_batched (Gee.ArrayList<Objects.Item> items) {
+        var related_items = new Gee.ArrayList<Objects.Item> ();
+
         foreach (var item in items) {
             string? parent_id = Util.find_string_value ("RELATED-TO", item.calendar_data);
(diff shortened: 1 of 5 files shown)


alainm23:
I’m not sure, but when syncing a Nextcloud account I’m getting this error.

```
ERROR:arraylist.vala:172:gee_array_list_real_get: assertion failed: (index < _size)
Bail out! ERROR:arraylist.vala:172:gee_array_list_real_get: assertion failed: (index < _size)
[1]    185407 IOT instruction (core dumped)  ./src/io.github.alainm23.planify --lang=en
```

core/Objects/Project.vala
```diff
@@ -531,6 +531,23 @@ public class Objects.Project : Objects.BaseObject {
         }
     }
 
+    public void add_items_batched (Gee.ArrayList<Objects.Item> items) {
```
alainm23, line 534: Collect items to remove in a separate list during iteration

```
public void add_items_batched (Gee.ArrayList<Objects.Item> items) {
    var items_to_remove = new Gee.ArrayList<Objects.Item> ();
    
    foreach (var item in items) {
        string? parent_id = Util.find_string_value ("RELATED-TO", item.calendar_data);
        
        if (parent_id != null && parent_id != "") {
            Objects.Item ? parent_item = Services.Store.instance ().get_item (parent_id);
            if (parent_item != null) {
                parent_item.add_item_if_not_exists (item);
                items_to_remove.add (item);
            }
        }
    }
    
    foreach (var item in items_to_remove) {
        items.remove (item);
    }
    
    if (!Services.Store.instance ().insert_items_transaction (items)) {
        error ("Failed to insert items in transaction");
    }
}
```

alainm23 (changes requested):

core/Objects/Project.vala
byquanton: I didn't check it with subtasks. I'll update it.

core/Services/CalDAV/CalDAVClient.vala
```diff
@@ -299,51 +301,51 @@ public class Services.CalDAV.CalDAVClient : Services.CalDAV.WebDAVClient {
...
-                foreach (var propstat in response.propstats ()) {
-                    if (propstat.status != Soup.Status.OK) {
-                        continue;
-                    }
+        int index = 0;
+        var items_list = new Gee.ArrayList<Objects.Item> ();
 
-                    var calendar_data = propstat.get_first_prop_with_tagname ("calendar-data");
-                    string? parent_id = Util.find_string_value ("RELATED-TO", calendar_data.text_content);
+        Idle.add (() => {
```
alainm23, line 310: Add bounds check at the start of Idle.add() callback to prevent race condition crash. The callback was accessing `responses[index]` before verifying `index < responses.size`, causing out-of-bounds errors when the loop completed. Moving the check before array access fixes the crash

This error occurs when a task list has zero items

```suggestion
                Idle.add (() => {
            if (index >= responses.size) {
                if (progress_callback != null) {
                    progress_callback (responses.size, responses.size, _ ("Loaded tasks for %s…").printf (project.name));
                }
                project.add_items_batched (items_list);
                Idle.add ((owned) callback);
                return false;
            }
```

alainm23 (changes requested):

PR #2185, opened 2026-01-11
Run translation updates in pull requests

byquanton:
Instead of adding commits to master after every PR that needs updates to the PO files. This PR changes it to run in the specifc pull request (after sending a message and asking)


.github/workflows/check-translations.yml
```diff
@@ -0,0 +1,92 @@
...
+                    fi
+
+            -   name: Comment on PR if updates are needed
+                if: steps.diff.outputs.changed == 'true'
+                uses: actions/github-script@v8
+                with:
+                    script: |
+                        const body = `
+                          **Translation files need updating**
+                          - [ ] **Apply translation updates**
```
alainm23, line 64:
```suggestion
                          - [x] **Apply translation updates**
```
Ideally, I’d prefer “Apply translation updates” to be enabled by default, and only disabled explicitly when needed. This keeps the workflow automatic and avoids relying on human memory

.github/workflows/apply-translations.yml
```diff
@@ -0,0 +1,71 @@
...
+                    
+                    python3 scripts/update_translations.py
+
+            -   name: Commit changes
+                run: |
+                    git add po/
+                    if ! git diff --staged --quiet; then
+                      git commit -m "Update translations"
+                      git push
+                    fi
```
alainm23, line 71: One more thing: with this flow we’d be missing the Weblate sync step.
In the current setup, Weblate is notified whenever translations are updated, which keeps everything in sync automatically.

I think we should also add a “Notify Weblate of changes” step after applying translation updates, otherwise Weblate won’t pull the latest changes

https://github.com/alainm23/planify/blob/master/.github/workflows/merge.yml#L82C9-L82C40

alainm23 (changes requested):

alainm23:
Thanks for this PR — it’s quite interesting and solves one of my issues with the current translation workflow, which is having to wait for translations to be automatically updated on the master branch. This fixes that and also helps keep the commit history cleaner. I just have a few small comments; otherwise, everything looks good

.github/workflows/apply-translations.yml
byquanton: Fixed

.github/workflows/check-translations.yml
byquanton: Automatic commits make sense for Weblate pull requests, but for normal PRs that might undergo merges or rebases while still in development, I think it's better to use a 'one-click' solution to add the commit.

PR #2207, opened 2026-01-26
Migrate to Adwaita 1.8 Shortcut Dialog

byquanton:


byquanton:
Todo:
- [ ] Fix Caldav integration branch to with newer libadwaita version

byquanton:
Should Planify remain compatible with libadwaita 1.7, which is used in Debian 13? If so, we could create a native DEB package. Then I would make the changes in this PR conditional on the libadwaita version present. Or should we move to the newer version?

alainm23:
I like the idea of using the most recent version more. It’s true that Planify has a larger user base, but I’d like to continue using Flatpak as the only distribution format

nicoroeser:
While preparing this pull request, please remember to add a commit which updates `README.md`:

1. update the minimum libadwaita version,
2. add blueprint-compiler as a development dependency.

Thanks!

PR #2226, opened 2026-02-01
Fix filter by labels in non project views

### LINKED ISSUE #2208: "Filter by Label" is not working in non project views
**Describe the bug**
I am using CalDav sync with a self-hosted server, and although I can create new labels and add them to tasks, I don't see a way to create new filters so I can add them to project views. For instance, if I go to the drop down in the top right of today view and click "Filter by labels", I am then presented with this empty menu that says "Your list of filters will show up here" with no option to add more filters.

**Expected behavior**
I would expect there to be a button to create a new filter in this menu, or at least a checkbox list like the filter by priority menu where I can check and uncheck labels I want to show up.

**Screenshots**
An image of the filter menu with the text cancelled in the search bar and the text Your list of filters will show up here in the results view

 - OS or DE: [e.g. pantheon, unity, gnome]
 - Version [e.g. 0.1.2]
 Gnome 49, EndeavourOS
Version 4.17.0

byquanton:
Fixes #2208.

And Refactored duplicated logic from the different Scheduled* classes into ScheduledSection.

### DIFF
--- a/po/POTFILES
+++ b/po/POTFILES
@@ -190,6 +190,7 @@ src/Views/Scheduled/ScheduledDay.vala
 src/Views/Scheduled/ScheduledMonth.vala
 src/Views/Scheduled/ScheduledOverdue.vala
 src/Views/Scheduled/ScheduledRange.vala
+src/Views/Scheduled/ScheduledSection.vala
 src/Views/Today.vala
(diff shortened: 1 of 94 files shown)


alainm23:
Great job, it looks really good, thank you very much.

PR #2240, opened 2026-02-07
Fix drag and drop after multi-select

### LINKED ISSUE #2116: Drag&Drop cannot be started after exiting multi-select mode
Tasks cannot be dragged anymore after the multi-selection mode has been used. This persists across switching lists etc. and is only fixed by restarting the app.

**To Reproduce**
1. Select multiple tasks via CTRL + click
2. Exit this mode via "Done" button
3. Try to drag any task (nothing will happen, the drag does not start)

**Expected behavior**
Drag and drop should work normally after the multi-selection stopped.

(Ideally, it should also work during that mode, but that's basically feature request #1431 )

**Desktop:**
 - OS or DE: Fedora 43, KDE Plasma
 - Version 4.16.1

pgrit:
Fixes #2116 

Calling `disable_drag_and_drop()` sets `drag_enabled` to false, 
https://github.com/alainm23/planify/blob/0b4eb0ae6530adc6ac8d79cd4923126f0632614f/src/Layouts/ItemRow.vala?plain=1#L1883-L1884

so drag and drop will never be re-enabled again because of this check:
https://github.com/alainm23/planify/blob/0b4eb0ae6530adc6ac8d79cd4923126f0632614f/src/Layouts/ItemRow.vala?plain=1#L835-L837

To fix this, I copied the existing logic from the `edit` setter 
https://github.com/alainm23/planify/blob/0b4eb0ae6530adc6ac8d79cd4923126f0632614f/src/Layouts/ItemRow.vala?plain=1#L145-L148

### DIFF
--- a/src/Layouts/ItemRow.vala
+++ b/src/Layouts/ItemRow.vala
@@ -824,7 +824,10 @@ public class Layouts.ItemRow : Layouts.ItemBase {
             if (active) {
                 select_revealer.reveal_child = true;
                 checked_button.sensitive = false;
-                disable_drag_and_drop ();
+                if (drag_source != null) {
+                    itemrow_box.remove_controller (drag_source);
+                    drag_source = null;
+                }
             } else {
                 select_revealer.reveal_child = false;
                 checked_button.sensitive = true;


src/Layouts/ItemRow.vala
```diff
@@ -824,7 +824,10 @@ public class Layouts.ItemRow : Layouts.ItemBase {
             if (active) {
                 select_revealer.reveal_child = true;
                 checked_button.sensitive = false;
-                disable_drag_and_drop ();
```
byquanton, line 827: Is the method `disable_drag_and_drop ()` used anywhere else?

pgrit: It's called in the "today" and "filter" views because those don't support reordering.

PR #2266, opened 2026-02-14
Update pt_PT.po

rottenpants466:
Updated Português Portugal translation.

Please check line 2254 to see if it is correct since it became too long.

### DIFF
--- a/po/pt_PT.po
+++ b/po/pt_PT.po
@@ -1186,7 +1186,7 @@ msgstr "Próxima semana"
 
 #: core/Widgets/DateTimePicker/DateTimePicker.vala:523
 msgid "Save"
-msgstr ""
+msgstr "Guardar"
 
 #: core/Widgets/DateTimePicker/DateTimePicker.vala:559
 msgid "Daily"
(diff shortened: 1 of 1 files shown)


alainm23:
Hi, while translating directly in the .po file is valid, the recommended approach is to do it through [Weblate](https://hosted.weblate.org/projects/planner/io-github-alainm23-planify/pt_PT/).
The action is currently failing with an error.

po/pt_PT.po
```diff
@@ -2247,11 +2247,12 @@ msgstr ""
 
 #: src/Dialogs/Preferences/Pages/QuickAdd.vala:103
 msgid "Keep Task Properties"
-msgstr ""
+msgstr "Manter as proriedades de Tarefas"
 
 #: src/Dialogs/Preferences/Pages/QuickAdd.vala:104
 msgid "When 'Keep adding' is enabled, preserve date, labels, and pinned status"
-msgstr ""
+msgstr "Quando o 'Continuar a adicionar' estiver ativado, preservar datas, etiquetas e "
```
alainm23, line 2254: Only a msgstr should go here — I think it needs to be something like this.

```
msgstr ""
"Quando o 'Continuar a adicionar' estiver ativado, preservar datas, etiquetas e  "
"estado do fixar"
```

alainm23 (changes requested):

po/pt_PT.po
rottenpants466: Ok, I will make those changes.

Sorry I forgot to make the changes in weblate. 

I will close this pull request.

Thank you

PR #2268, opened 2026-02-14
CalendarEvents: Avoid Posix.NLTime if not available

fossdd:
_NL_TIME_FIRST_WEEKDAY and similar are glibc-ism's and don't exist on other libc implementations like musl libc.

So if it's not available, set first weekday to 0 instead.


alainm23:
@fossdd Hi, and thanks for contributing to Planify. I have a question — what exactly does this change do?

fossdd:
It checks per meson if _NL_TIME_FIRST_WEEKDAY is available, then passes that as to the vala preprocessor (https://gnome.pages.gitlab.gnome.org/vala/manual/preprocessor.html) as symbol with valac -D, and then either uses Posix.NLTime if it's available, or if not, just use 0 as the first weekday.

or what else is not understandable about the patch? :)

PR #2324, opened 2026-03-27
Add PDF export for projects

### LINKED ISSUE #552: Possibility to print project/tasks
I was surprised by not being able to print my current project as a pdf or on a connected printer
So this is a problem to me regarding the app.

We should be allowed to print our project/task as per need or request by team members
We should be able to print :
  - per project (full content)
  - Per tasks
  - a general/ surface view of a project (not including details here)
in the most common exchange file format like PDF etc.

we also should be allowed to share to Telegram, Element, Discord, mail etc. as this may be requested and would help us not loose time but be more efficient.

alainm23:
Adds the ability to export a project as a PDF file from the project context menu ("Export as PDF").

The generated PDF includes:

- Project name, description, and deadline
- Tasks with status, due date, priority (colored), and deadline
- Sections ordered by position with their tasks
- Sub-tasks with indentation
- Automatic page breaks

Uses Cairo.PdfSurface directly, no extra dependencies needed.

Fixes: #552

### DIFF
--- a/core/Widgets/IconColorProject.vala
+++ b/core/Widgets/IconColorProject.vala
@@ -26,7 +26,6 @@ public class Widgets.IconColorProject : Adw.Bin {
     private Widgets.CircularProgressBar circular_progress_bar;
     public Gtk.Image inbox_icon;
     private Gtk.Label emoji_label;
-    private Gtk.Stack color_emoji_stack;
     private Gtk.Stack stack;
 
     private Gee.HashMap<ulong, weak GLib.Object> signal_map = new Gee.HashMap<ulong, weak GLib.Object> ();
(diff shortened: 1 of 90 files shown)


RokeJulianLockhart:
@alainm23, how is it represented within the PDF – as in, as a bitmap, or vector? I ask because I'd like any export to be lossless and small, if possible.

alainm23:
@RokeJulianLockhart  The PDF export uses `Cairo.PdfSurface` which renders everything as vector graphics — text, lines, and shapes are all stored as vectors, not bitmaps. This means the output is lossless, resolution-independent, and typically very small in file size.

Issue #2127, opened 2025-12-13
Update due date labels after waking from sleep


pgrit:
Currently, if I put my PC to sleep over night -- with planify staying open -- then all displayed due dates will be inaccurate the next morning. E.g., tasks due "today" will still display the label for due "tomorrow"

<img width="214" height="99" alt="Image" src="https://github.com/user-attachments/assets/75b7f475-6ad3-4669-b95e-d6ffb6481dd7" />

An automatic update would be nice. Alternatively, maybe the "refresh" button could at least update these labels, because currently the only solution seems to be to restart planify.

PR #2362, opened 2026-04-14
Fix day change not detected after system suspend/resume

### LINKED ISSUE #2173: Reschedule "Today" schedules the day when Planify was started
**Describe the bug**
In Today view, tasks with a due date in the past are correctly placed in category "Overdue".  
Rescheduling to "Today" is not working because it's always scheduled to the day when Planify was started.

**To Reproduce**
1. Ensure tasks with due date today
2. Wait until midnight (keep Planify open)
3. The tasks are displayed as "Overdue"
4. Click "Reschedule" » "Today"
5. all overdue tasks are rescheduled to the date when Planify was opened

**Expected behavior**
The date for "Today" needs to be updated to be correct over the time.

**Screenshots**
Image

You can see the date at "Choose a date" is "Jan 3", but today is "Jan 5".  
I had overdue tasks from "Jan 3" and "Jan 4" and after clicking "Reschedule > Today" all tasks were scheduled to "Jan 3" instead of "Jan 5".

 - Kubuntu 24.04, KDE
 - Planify 4.17.0

alainm23:
Replace 5-minute polling with precise midnight scheduling that fires exactly at day change. On resume from suspend, the timeout is cancelled and rescheduled from the current time to avoid stale timers.

Fixes: #2173 #2127

### DIFF
--- a/build-aux/io.github.alainm23.planify.Devel.json
+++ b/build-aux/io.github.alainm23.planify.Devel.json
@@ -15,7 +15,6 @@
         "--socket=wayland",
         "--talk-name=org.gnome.evolution.dataserver.Calendar8",
         "--talk-name=org.gnome.evolution.dataserver.Sources5",
-        "--system-talk-name=org.freedesktop.login1",
         "--talk-name=org.gnome.ScreenSaver",
         "--talk-name=io.github.alainm23.planify",
         "--talk-name=io.github.alainm23.planify.quick-add",
--- a/build-aux/io.github.alainm23.planify.Devel.json
+++ b/build-aux/io.github.alainm23.planify.Devel.json
@@ -15,6 +15,8 @@
         "--socket=wayland",
         "--talk-name=org.gnome.evolution.dataserver.Calendar8",
         "--talk-name=org.gnome.evolution.dataserver.Sources5",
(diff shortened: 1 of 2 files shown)


alexreisner:
@alainm23 thank you for the great app! There still seems to be a slight issue here, in one very specific case. I can easily work around it but I think new users of the app might be confused. Steps to reproduce:

1. In the evening, put Planify in "Today" view and close laptop lid.
2. Open it in the morning. Yesterday's tasks are (correctly) listed under "Overdue," but when I click the Reschedule button and then the Today button, nothing happens. The date for each task is still "Yesterday."

In calendar popups, the app does highlight the correct date, and I can click that to reschedule, after which, the "Today" button seems to work when rescheduling or creating any other tasks. So I think this problem only occurs in the one specific case: clicking the "Today" button under "Reschedule."

I'm happy to provide more details on reproducing this if those steps don't work for you.

alexreisner:
@alainm23 to add a bit more detail here: it seems that the Today and Yesterday buttons in the reschedule popup are still tied to the date the app was originally opened. If I've had the app open for 5 days and click Reschedule > Tomorrow, it changes the date on all overdue tasks to 4 days ago.

PR #2397, opened 2026-04-22
Fix CalDAV sync for multiple Server Implementations

### LINKED ISSUE #2163: CalDAV sync works only 1 time, not fully thereafter
After adding a new CalDAV account successfully, the tasks of 1 out of my 2 calendars there is synced. For the second calendar the sync circle goes round and round. For any other sync there is the warning triangle saying "connection to server failed". 

However, the connection seems to be up: 
When adding or changing a task in Planify, this is still synced to the server
On the other hand, a task added or changed in a different client on the smartphone does not reach Planify. (No problem with the other client: when deleting and re-adding the account in Planify, then also the added/updated task is visible in Planify after initial sync).
So: outgoing sync works, incoming sync fails, error "connection to server failed" is not (completely) true. 

**Expected behavior**
outgoing sync works, incoming sync works 

 - OS or DE: Linux Mint 22.2 XFCE
 - 7.14.0 (flatpak)

byquanton:
Makes it possible to sync via caldav if the server doesn't support the sync token.

I extracted the 'upsert_vtodo_content' code from the 'sync_tasklist' method in order to create a generic implementation shared by full and partial syncs.

I finally created a Posteo account to test this on a server with such an implementation and fix the issue.

Furthermore, I don't think #2368 fixed anything. This is because DAV usually returns a multi-status response, and these always have a 207 status code. Have you tested the PR with a CalDAV server that doesn't support it? If not, I think the commits should be reverted.

Fixes #2163

### DIFF
--- a/core/Services/CalDAV/CalDAVClient.vala
+++ b/core/Services/CalDAV/CalDAVClient.vala
@@ -394,8 +394,13 @@ public class Services.CalDAV.CalDAVClient : Services.CalDAV.WebDAVClient {
 
         yield fetch_project_details (project, cancellable);
 
+        Services.LogService.get_default ().debug ("CalDAV", "sync_id after fetch_project_details: '%s'".printf (project.sync_id ?? "(null)"));
+
         if (project.sync_id == null || project.sync_id == "") {
+            Services.LogService.get_default ().warn ("CalDAV", "No sync-token from server, falling back to etag-based sync for '%s'".printf (project.name));
             project.loading = false;
+            yield etag_sync_project (project, cancellable);
+            project.sync_finished ();
             return;
         }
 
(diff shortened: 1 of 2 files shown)


byquanton:
_pressed the wrong button_

The branch works in its current state, but I'm looking into how I can refactor it to use 'add_items_batched' or, rather, how I can make it work with batched inserts.

alainm23:
Thanks for the improvements. I ran some tests for PR #2368 using a Vikunja server in Docker just for experimentation, and the solution does work. However, I didn’t test it using Posteo — it might have a different configuration. I’ll try to set up a local environment as well to test it.

alainm23:
I've been working on the same issue and just opened a PR with a fix — would 
love if you could take a look and share your thoughts! The approach I went with 
is an ETag-based sync fallback: when the server doesn't return a `sync-token`, 
it fetches only the ETags first (lightweight REPORT), then downloads the full 
VTODO only for changed or new items, and deletes local items that are no longer 
on the server.

Really appreciate the effort you put into this 🙏

byquanton:
> Thanks for the improvements. I ran some tests for PR https://github.com/alainm23/planify/pull/2368 using a Vikunja server in Docker just for experimentation, and the solution does work. However, I didn’t test it using Posteo — it might have a different configuration. I’ll try to set up a local environment as well to test it.

The current release (4.19.0) has some issues by using fetch_items_for_project, because it is only able to add items, not really update or remove ones. Also, the item counter is broken and updates on every sync.

https://github.com/user-attachments/assets/bf9e1a6f-4105-495b-a8de-730706365496

> I've been working on the same issue and just opened a PR with a fix — would
love if you could take a look and share your thoughts! The approach I went with
is an ETag-based sync fallback: when the server doesn't return a sync-token,
it fetches only the ETags first (lightweight REPORT), then downloads the full
VTODO only for changed or new items, and deletes local items that are no longer
on the server.

I've rebased and refactored this pull request (PR) on top of your changes. The second commit moves the upsert logic (adding and updating) into a single method shared by the three sync implementations (initial sync, ETag sync, and sync-collection). I've also changed the catch that was used to 'fix' Vikunja to use etag sync, as I noticed no issues with this approach. If necessary, changing it back to project sync would only require a single line change, since fetch_items_for_project now also uses the new upsert logic.

PR #2416, opened 2026-04-28
Add support for system accent color instead of hardcoded accent color

Dhanush-Proj:
Used  Adw.StyleManager APIs to get the user's GNOME accent color. Falls back to #3584e4 (blue) if unsupported.


alainm23:
The feature is valid, but I’d like to experiment with it a bit more — maybe with an additional setting indicating that it can use the system’s default accent color. I say this because Planify’s brand color has always been blue. Still, it’s good work, approved for now.

alainm23 (approved):

Dhanush-Projectile:
In that case how  about  a toggle button in appearance for using accent color

PR #2423, opened 2026-04-29
Fix Caldav sync-collection sync

### LINKED ISSUE #2422: Spaces in XML/Hangs while syncing with NextCloud server
### Prerequisites

- [x] I have searched the [existing issues](https://github.com/alainm23/planify/issues) and this bug has not been reported yet

### Describe the bug

Since upgrading to version 4.19.0, every time I try to sync with my NextCloud account, the app freezes for a few seconds with high CPU usage while syncing each calendar, and, as a result, the synchronization is extremely slow. Rolling the app back to version 4.18.3 fixed the issue.

### Steps to reproduce

1. Install Pllanify 4.19.0 or 4.19.1
2. Add NextCloud account
3. Start synchronization

### Expected behavior

Fast synchronization with low CPU usage

### Installation method

Built from source

### Planify version

4.19.1

### Desktop environment

KDE Plasma, Sway

### Linux distribution

Arch Linux

### Sync provider (if relevant)

Nextcloud (CalDAV)

### Screenshots or logs

_No response_

byquanton:
Fixes #2422

### DIFF
--- a/core/Services/CalDAV/CalDAVClient.vala
+++ b/core/Services/CalDAV/CalDAVClient.vala
@@ -377,8 +377,7 @@ public class Services.CalDAV.CalDAVClient : Services.CalDAV.WebDAVClient {
 
         Services.LogService.get_default ().info ("CalDAV", "Syncing tasklist");
 
-        var xml = """
-        <?xml version='1.0' encoding='utf-8'?>
+        var xml = """<?xml version='1.0' encoding='utf-8'?>
         <d:sync-collection xmlns:d="DAV:">
             <d:sync-token>%s</d:sync-token>
             <d:sync-level>1</d:sync-level>


src/MainWindow.vala
```diff
@@ -214,12 +214,6 @@ public class MainWindow : Adw.ApplicationWindow {
                 return GLib.Source.REMOVE;
             });
 
-            Services.Store.instance ().source_added.connect ((source) => {
```
alainm23, line 217: The reason I added this is so that when a source is added, `run_server` is enabled by default and updates happen automatically. Maybe the ideal approach isn’t to remove it, but instead to add a flag so it doesn’t sync the first time and does it in the next iteration.

alainm23 (changes requested):

src/MainWindow.vala
byquanton: Maybe we can move that logic to the set of sync_server, so just calling source.sync_server = true is enough?

byquanton: I've dropped that commit. I'll create a separate PR for it.

alainm23 (approved):

PR #2447, opened 2026-05-09
Toggle under a group looks better now 

Dhanush-Proj:
Previous error was from my side produced by my setup Indentation issue . So this was a Toggle under Accent color group in appearance setting as a separate group named Accent Color   for consistency

### DIFF
--- a/core/Utils/Util.vala
+++ b/core/Utils/Util.vala
@@ -279,13 +279,15 @@ public class Util : GLib.Object {
         }
 
         string accent_color = "#3584e4";
-        if (Adw.StyleManager.get_default ().get_system_supports_accent_colors ()) {
-            var rgba = Adw.StyleManager.get_default ().get_accent_color_rgba ();
-            accent_color = "#%02x%02x%02x".printf (
-                (uint) (rgba.red * 255),
-                (uint) (rgba.green * 255),
-                (uint) (rgba.blue * 255)
-            );
+        if (Services.Settings.get_default ().settings.get_boolean ("use-system-accent")) {
+            if (Adw.StyleManager.get_default ().get_system_supports_accent_colors ()) {
+                var rgba = Adw.StyleManager.get_default ().get_accent_color_rgba ();
+                accent_color = "#%02x%02x%02x".printf (
+                    (uint) (rgba.red * 255),
+                    (uint) (rgba.green * 255),
+                    (uint) (rgba.blue * 255)
+                );
+            }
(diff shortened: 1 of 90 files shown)


core/Utils/Util.vala
```diff
@@ -307,8 +309,8 @@ public class Util : GLib.Object {
                 upcoming_fg_color = "#f0f0f0";
                 selected_color = "#2e3a46";
                 card_bg_color = "#222222";
-                Adw.StyleManager.get_default ().color_scheme = Adw.ColorScheme.FORCE_DARK;
-            } else if (appearance_mode == Appearance.DARK_BLUE) {
+                Adw.StyleManager.get_default ().color_scheme = Adw.ColorScheme.PREFER_DARK;
```
alainm23: One small thing I noticed, there's a change from FORCE_DARK to PREFER_DARK in the dark mode section that seems unrelated to the accent color feature. Was that intentional? Just want to make sure it doesn't cause any unexpected behavior for users who have dark mode enabled.

src/Dialogs/Preferences/Pages/Appearance.vala
```diff
@@ -57,11 +59,27 @@ public class Dialogs.Preferences.Pages.Appearance : Dialogs.Preferences.Pages.Ba
         system_appearance_row.set_activatable_widget (system_appearance_switch);
         system_appearance_row.add_suffix (system_appearance_switch);
 
+        system_accent_switch = new Gtk.Switch () {
```
alainm23, line 62: The "Use System Accent Color" toggle is always visible even on systems that don't support accent colors. In that case the toggle wouldn't really do anything, which might confuse users. Would be nice to hide it when the system doesn't support it! 😄

alainm23 (changes requested):

alainm23:
Hey, thanks for the PR! The system accent color toggle is a nice addition.

core/Utils/Util.vala
Dhanush-Projectile: i experimented on it i'm sorry it got pushed

src/Dialogs/Preferences/Pages/Appearance.vala
Dhanush-Projectile: Right lemme fix that

alainm23 (approved):

Dhanush-Projectile:
Thank you

Issue #1920, opened 2025-10-12
Neither the storage location of a backup is indicated, nor can it be chosen by the user


4jNsY6fCVqZv:
I would like to see the user interface offer me both.

4.15.1 (Flatpak)

tredondo:
I tried to click the triple dot menu next to a backup, then Download, and I was offered to download a JSON file, so this isn't that bad. But,

1. The second time I attempted this a few seconds later, Planify crashed.

<details><summary>Logs</summary>
** (io.github.alainm23.planify:2): CRITICAL **: 22:24:20.665: objects_source_data_to_json: assertion 'self != NULL' failed

(io.github.alainm23.planify:2): Json-CRITICAL **: 22:24:20.667: json_builder_end_object: assertion 'json_builder_current_mode (builder) == JSON_BUILDER_MODE_OBJECT' failed

(io.github.alainm23.planify:2): Json-CRITICAL **: 22:24:20.667: json_builder_end_object: assertion 'json_builder_current_mode (builder) == JSON_BUILDER_MODE_OBJECT' failed

(io.github.alainm23.planify:2): Json-CRITICAL **: 22:24:20.667: json_builder_end_object: assertion 'json_builder_current_mode (builder) == JSON_BUILDER_MODE_OBJECT' failed

(io.github.alainm23.planify:2): Json-CRITICAL **: 22:24:20.667: json_builder_end_object: assertion 'json_builder_current_mode (builder) == JSON_BUILDER_MODE_OBJECT' failed

(io.github.alainm23.planify:2): Json-CRITICAL **: 22:24:20.667: json_builder_end_object: assertion 'json_builder_current_mode (builder) == JSON_BUILDER_MODE_OBJECT' failed

(io.github.alainm23.planify:2): Json-CRITICAL **: 22:24:20.667: json_builder_end_object: assertion 'json_builder_current_mode (builder) == JSON_BUILDER_MODE_OBJECT' failed

(io.github.alainm23.planify:2): Json-CRITICAL **: 22:24:20.667: json_builder_end_object: assertion 'json_builder_current_mode (builder) == JSON_BUILDER_MODE_OBJECT' failed

(io.github.alainm23.planify:2): Json-CRITICAL **: 22:24:20.667: json_builder_end_object: assertion 'json_builder_current_mode (builder) == JSON_BUILDER_MODE_OBJECT' failed

[...]

PR #2459, opened 2026-05-13
feat: automatic daily backup with extra folder support

### LINKED ISSUE #2019: Add option to schedule backups
This way, my Planify database could become a regular part of my scheduled system backups without having to manually create a Planify backup each time before.

Planify 4.15.2 (Flatpak)

alainm23:
Adds automatic daily backup support to Planify. Backups now trigger automatically at midnight and can be copied to additional folders configured by the user. Also includes several fixes to the backup export/import data integrity and improvements to the Restore Backup UI.

Fixes: #2019 #1920

image


4jNsY6fCVqZv:
A big thank you!
What happens if my device is switched off at midnight?
And would a customisable time setting be useful in that case?

alainm23:
Hi,

Great question! If your device is off at midnight, the backup won't run at that exact moment — but don't worry. When you turn your device back on and open Planify, the day change is detected and the automatic backup will trigger then.

As for a customisable time setting, we don't think it would add much value in practice since the backup is tied to the day change rather than a specific clock time. The goal is simply to ensure one backup per day happens, whenever the app is running.

Thanks for the kind words and for thinking about edge cases like this!

4jNsY6fCVqZv:
> . The goal is simply to ensure one backup per day happens, whenever the app is running.

That works fine for me.
Technically speaking, does the missed backup get rescheduled or is it simply skipped in favour of the backup for the current day?

> Thanks for the kind words and for thinking about edge cases like this!

With pleasure!

4jNsY6fCVqZv:
I installed the update with the backup feature on the very first day.
And here’s what I’ve noticed: Unfortunately, there’s only been a single backup so far.
That was when I had Planfy open and running in the foreground exactly at midnight.
On all the other days when I had Planfy closed or my laptop switched off, there was no daily backup.
Nor was there a catch-up backup when I opened the app again.

PR #2461, opened 2026-05-13
Support builds with libical 4

achmafooma:
libical version 4 removed the `get_by_day_array` and `set_by_day_array` methods, which were replaced with `get_by_array` and `set_by_array` with an `ICal.RecurrenceByRule.BY_DAY` parameter.

See libical documentation: [Migrating to version 4](https://github.com/libical/libical/blob/4.0/docs/MigrationGuide_to_4.md)

Arch-based Linux distributions have already switched to libical 4, which broke new builds. But other distros (and Windows MSYS2) are still using libical 3.

This change adds conditional logic so the correct method is used whether it's building against libical 3 or 4.

### DIFF
--- a/core/Objects/Item.vala
+++ b/core/Objects/Item.vala
@@ -1319,7 +1319,11 @@ public class Objects.Item : Objects.BaseObject {
                         }
                     }
 
+                    #if IS_LIBICAL4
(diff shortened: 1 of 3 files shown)


byquanton (commented):
The PR looks good. Just one thing: can you remove the unnecessary whitespace changes in Item.vala?

alainm23 (approved):

PR #2473, opened 2026-05-15
fix: cursor no longer jumps to end when editing tasks in Today and Scheduled views

### LINKED ISSUE #2220: Cursor jumps to end of description
**Describe the bug**
A few seconds after editing the description of a task, the cursor jumps to the end of the description.

**To Reproduce**
1. Create a task.
2. Type something in the description.
3. Quickly, using either the mouse or keyboard, move the text cursor somewhere earlier in the description.
4. After a few seconds, the text cursor will jump to the end of the description.

**Expected behavior**
The cursor should remain where the user places it.

**Screenshots**
N/A

 - OS or DE: Fedora 43, Plasma 6.5.5 with Wayland
 - Planify Version: 4.17.0

**Additional context**
N/A

alainm23:
When editing a task title or description in the Today or Scheduled views, the cursor would unexpectedly jump to the end of the text a few seconds after stopping to type. This made it frustrating to edit text in the middle of a task.

The root cause was that these views were reloading the full content of the task widget every time an update was saved, even when the update originated from that same widget. The fix applies the same pattern already used in the Project view — skipping the reload when the update came from the widget itself, so the cursor stays exactly where the user left it.

Fixes: #2220

### DIFF
--- a/src/Views/Scheduled/ScheduledDay.vala
+++ b/src/Views/Scheduled/ScheduledDay.vala
@@ -157,7 +157,7 @@ public class Views.Scheduled.ScheduledDay : Views.Scheduled.ScheduledSection {
     }
 
     protected override void valid_update_item (Objects.Item item, string? update_id = null) {
-        if (items.has_key (item.id)) {
+        if (items.has_key (item.id) && items[item.id].update_id != update_id) {
             items[item.id].update_request ();
         }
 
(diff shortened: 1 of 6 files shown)


4jNsY6fCVqZv:
Thanks for taking the time to look at it again. And thanks for your explanation, too.

PR #2520, opened 2026-05-29
add zh_CN.po

blusewang:
add chinese translate po file

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -1,0 +1,3740 @@
+msgid ""
+msgstr ""
+"Language: zh_CN \n"
+"Language-Team: Chinese (Simplified Han script)\n"
+"Last-Translator: jeff <jeff@wangjunfeng.com.cn>\n"
+
+#: core/Enum.vala:168 src/Dialogs/Preferences/Pages/Backup.vala:511
+#: src/Dialogs/Preferences/Pages/General.vala:55
+msgid "Projects"
+msgstr "项目"
+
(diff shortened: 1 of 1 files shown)


po/zh_CN.po
```diff
@@ -0,0 +1,3740 @@
+msgid ""
+msgstr ""
+"Language: zh_CN \n"
```
alainm23, line 3: Thanks for the PR, it looks good. Just one small comment: you still need to add the `zh_CN` language to `po/LINGUAS`, otherwise Planify won’t know that the language exists.

alainm23 (changes requested):

byquanton:
Duplicates the changes from #2521 into a new file?

Both are Chinese (Simplified Han script)?

blusewang:
> Duplicates the changes from #2521 into a new file?
> 
> Both are Chinese (Simplified Han script)?

yes! without zh_CN.po Planify can work fine with Chinese (Simplified Han script). Submitting zh_CN.po is to fully implement the Simplified Chinese translation section in the po/README.md document.

byquanton:
What is the difference to #2521? 

Isn't just updating zh.po enough?

If your locale is set to zh_CN it should fall back to zh?

byquanton:
If both have the same content, a duplicate file isn't required.

blusewang:
Yes, you're right. I'll close this PR and switch to updating zh.po.

PR #2521, opened 2026-05-29
complate zh.po translate

blusewang:


po/zh.po
```diff
@@ -1,18 +1,8 @@
 msgid ""
```
alainm23, line 1: Thanks for the translation! The CI is failing because the zh.po header was modified and is now missing required fields. Please restore the original header and only update the translation strings below it.

The header should look like this:

```
msgid ""
msgstr ""
"Project-Id-Version: io.github.alainm23.planify\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Language: zh\n"
```

alainm23 (changes requested):

po/zh.po
blusewang: It fixed

PR #2522, opened 2026-05-29
update zh.po header

blusewang:


byquanton:
why does your pr just tries to delete the header?

PR #2530, opened 2026-06-13
Fix todoist recurring tasks

### LINKED ISSUE #1584: Planify doesn't seem to read repeating tasks from Todoist's API properly
**Describe the bug**
Lately, when I try to check off a completed task in my Todoist Inbox (or any Inbox, Today, Scheduled, etc. items that come from Todoist), Planify doesn't succeed in checking off these items.

For example, let's say I have a daily Todoist task called "Taking out the trash". Planify will tell me that this task is due today, June 10. I complete the task and click on the box next to the "Taking out the trash" task. Planify will briefly put a checkmark in the box, but then the checkmark will disappear and the "Taking out the trash" task will remain in the same place in the list, and Planify will give a notification pop-up: "Completed. Next occurence: June 10". No matter how many times I try to check off the task, it will always remain on June 10 until I go into the Todoist website or my Todoist phone app and check it off there. After I've checked it off on the Todoist website or app, then Planify will update it to show the next "Taking out the trash" task on June 11 (since it's a daily repeating task).

I have tried removing all tasks and re-entering them to make sure they are properly entered. Todoist doesn't show any problems on their end.

**To Reproduce**
1. Create a task.
2. Check off the task in Planify.
3. The described behavior will occur.

**Expected behavior**
The checking off of tasks should remove them, or move them to their next occurrence date if they are set to repeat.

 - OS or DE: Linux Mint 22.1, Cinnamon desktop
 - Planify Version: 4.12.0 (installed from Flatpak/Flathub)

korintheninth:
This fixes #1584 and Todoist recurring tasks integration in general.
#### What was:
Recurrence information wasn't sent as Todoist expected a NL string for recurring tasks which wasn't generated.
#### Whats now:
Generates NL string in English that successful creates an equivalent recurring task at Todoist. uses item_close instead of item_complete or item_update as Todoist recommends and item_close automatically triggers a reschedule at Todoist
#### Whats still not fixed:
there are recurrence rules that Planify support and Todoist doesn't and vice versa.
#### What was:
Recurrence information was simply ignored despite chrono being able to parse NL into Planify properties.
#### Whats now:
App now stores the recurrence string, uses chrono to parse it and successfully generates an equivalent recurring task.
#### Whats still not fixed:
chrono is not perfect
#### What was:
Tasks with reminders that were created while offline would create zombie tasks that would return after deletion after each restart.
#### Whats now:
Added on update cascade to tables now they work properly. touched queue logic now retries on temp_id failures
#### Whats still now fixed:
Requires a database recreation to take effect.

#### AI disclosure:
This is my first time touching vala so i heavily relied on ai guidance and code may not be perfect. None of the code was directly ai generated but none of it is also fully free of it.

### DIFF
--- a/core/Objects/DueDate.vala
+++ b/core/Objects/DueDate.vala
@@ -21,6 +21,7 @@
 
 public class Objects.DueDate : GLib.Object {
     public string date { get; set; default = ""; }
+    public string recurrence_string { get; set; default = ""; }
     public string time_zone { get; set; default = ""; }
     public string recurrency_weeks { get; set; default = ""; }
     public bool is_recurring { get; set; default = false; }
@@ -103,6 +104,10 @@ public class Objects.DueDate : GLib.Object {
             date = object.get_string_member ("date");
         }
 
+        if (object.has_member ("string")) {
+            recurrence_string = object.get_string_member ("string");
+        }
+
         if (object.has_member ("timezone")) {
             time_zone = object.get_string_member ("timezone");
         }
@@ -117,6 +122,9 @@ public class Objects.DueDate : GLib.Object {
         if (object.has_member ("date")) {
             date = object.get_string_member ("date");
         }
+        
+        if (object.has_member ("recurrence_string"))
+            recurrence_string = object.get_string_member ("recurrence_string");
 
         if (object.has_member ("timezone")) {
             time_zone = object.get_string_member ("timezone");
@@ -149,6 +157,7 @@ public class Objects.DueDate : GLib.Object {
 
     public void reset () {
         date = "";
+        recurrence_string = "";
         time_zone = "";
         recurrency_type = RecurrencyType.NONE;
         recurrency_interval = 0;
@@ -161,6 +170,9 @@ public class Objects.DueDate : GLib.Object {
 
         builder.set_member_name ("date");
         builder.add_string_value (date);
+        
+        builder.set_member_name ("recurrence_string");
+        builder.add_string_value (recurrence_string);
 
         builder.set_member_name ("timezone");
         builder.add_string_value (time_zone);
@@ -208,6 +220,7 @@ public class Objects.DueDate : GLib.Object {
     public Objects.DueDate duplicate () {
         var new_due = new Objects.DueDate ();
         new_due.date = date;
+        new_due.recurrence_string = recurrence_string;
         new_due.time_zone = time_zone;
         new_due.recurrency_weeks = recurrency_weeks;
         new_due.is_recurring = is_recurring;
--- a/core/Objects/DueDate.vala
+++ b/core/Objects/DueDate.vala
@@ -123,8 +123,9 @@ public class Objects.DueDate : GLib.Object {
             date = object.get_string_member ("date");
         }
         
-        if (object.has_member ("recurrence_string"))
+        if (object.has_member ("recurrence_string")) {
             recurrence_string = object.get_string_member ("recurrence_string");
+        }
 
         if (object.has_member ("timezone")) {
             time_zone = object.get_string_member ("timezone");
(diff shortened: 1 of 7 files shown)


alainm23:
Sorry, I haven’t had much time to review this PR. My work has been consuming most of my time lately. I hope you can understand, and thank you for your patience

core/Utils/Datetime.vala
```diff
@@ -486,6 +516,76 @@ public class Utils.Datetime {
...
+    public static string due_to_todoist_natural_language (Objects.DueDate due) {
+        if (due.recurrence_string != "" && due.is_recurring) {
+            return due.recurrence_string;
+        }
+        
+        if (due.date != "" && !is_iso_due_date (due.date)) {
+            return due.date;
+        }
+
+        if (due.is_recurring == false || due.recurrency_type == 0 || due.recurrency_type == 6) {
```
alainm23: These magic numbers are fragile. Consider using` RecurrencyType.NONE` and the corresponding enum value for 6 instead of raw integers.

core/Utils/Datetime.vala
```diff
@@ -486,6 +516,76 @@ public class Utils.Datetime {
...
+            if (due.recurrency_weeks == "1,2,3,4,5") {
+                returned += "weekday";
+                if (due.recurrency_interval > 1) {
+                    returned += "s";
+                }
+                returned += " ";
+            } else {
+                var selected_days = due.recurrency_weeks.split (",");
+                foreach (var day in selected_days) {
+                    returned += days[int.parse (day) - 1];
```
alainm23: Potential index out of bounds if `day` is "0" or > "7". Might be worth adding a bounds check here.

core/Utils/Datetime.vala
```diff
@@ -486,6 +516,76 @@ public class Utils.Datetime {
...
+                }
+                returned = returned[0:-1] + " ";
+            }
+        }
+
+        if (has_time (due.datetime)) {
```
alainm23: If `due.datetime` is null (recurring task without a date set), this could be problematic. `has_time` does handle null, but the subsequent `due.datetime.format()` on line 573 would crash. Worth adding a null guard.

core/Objects/DueDate.vala
```diff
@@ -117,6 +122,9 @@ public class Objects.DueDate : GLib.Object {
         if (object.has_member ("date")) {
             date = object.get_string_member ("date");
         }
+        
+        if (object.has_member ("recurrence_string"))
```
alainm23: Minor style inconsistency: missing braces here while the rest of the file uses them. Not blocking but worth keeping consistent.

core/Services/Database.vala
```diff
@@ -308,7 +308,7 @@ public class Services.Database : GLib.Object {
                 due                 TEXT,
                 mm_offset           INTEGER,
                 is_deleted          INTEGER,
-                FOREIGN KEY (item_id) REFERENCES Items (id) ON DELETE CASCADE
+                FOREIGN KEY (item_id) REFERENCES Items (id) ON DELETE CASCADE ON UPDATE CASCADE
```
alainm23, line 311: As you mentioned, this won't apply retroactively on existing SQLite databases. Would be good to add a migration step (recreate affected tables) or at minimum document clearly in release notes that users need to reset their DB for this fix to take effect.

alainm23 (changes requested):

korintheninth:
Made the requested changes and added the migration function. Also fixed Todoist - Planify reschedule recurring task mismatch. Todoist reschedules to first recurrence after the time of completion while Planify reschedules to next recurrence. Now it gets the next due date from Todoist response and updates locally.

alainm23:
LGTM, Previous concerns addressed. NL generation, sync flow, and DB migration all look good. Approving.

alainm23 (approved):

PR #2654, opened 2026-08-29
feat: sort and filter options in All Tasks view

### LINKED ISSUE #2431: Sorting options for all task view
### Prerequisites

- [x] I have searched the [existing issues](https://github.com/alainm23/planify/issues) and this feature has not been requested yet

### Is your feature request related to a problem?

It is hard to get the overview about all tasks over all projects in order to pick the next best task to work on.

### Describe the solution you'd like

Add a **Sort** control to the All Tasks view toolbar that supports two levels of sorting — a primary key and a secondary key for tie-breaking.

The default sort should be:

```
Primary:   Due date    (ascending — soonest first)
Secondary: Priority    (descending — Urgent → High → Medium → Low → None)
```

Supported sort criteria should include at minimum:

- **Due date** — ascending or descending
- **Priority** — Urgent → High → Medium → Low → None (or reversed)
- **Label** — alphabetical by first assigned label
- **Task list / project** — alphabetical by parent list name
- **Title** — alphabetical
- **Creation date** — ascending or descending

### Alternatives you've considered

_No response_

### Additional context

_No response_

CR0CKER:
### What changed

The All Tasks view gains the sort and filter controls that project views already have:

- **Sorting** — Project, Alphabetically, Due Date, Date Added, Date Modified, Priority, with an
  ascending/descending toggle. The comparison itself delegates to the existing
  `Util.set_item_sort_func`, so ordering semantics stay identical to the Today and Scheduled views.
- **Filters** — due-date range, priority (P1–P4) and labels, evaluated by the existing
  `Utils.TaskUtils.items_filter_func` and stored in the `filters` bag `Objects.BaseObject` already
  provides. Behaviour matches the project view, including its OR-across-filter-types semantics.
- An indicator dot on the view-settings button when the view is not in its default state, matching
  `Views.Project`.
- **Filter chips** above the list, one per active filter, each removable — the same affordance the
  project, Today and Scheduled views already offer. This reuses `Widgets.FilterFlowBox` unchanged:
  the All Tasks filters already live in the `Objects.BaseObject` filter bag the widget reads, so it
  needs nothing but its `base_object`, and removing a chip routes through the same
  `remove_filter ()` path the menu uses, leaving the pickers in step.

No new widgets were introduced — the popover is built from the existing
`Widgets.ContextMenu.{MenuPicker, MenuCheckPicker, MenuSwitch, MenuItem, MenuSeparator}`, and the
label filter reuses `Dialogs.LabelPicker`.

### Why

Closes #2431. It is hard to get an overview across projects to pick what to work on next, and the
view offered no ordering control at all.

The scope deliberately follows your comment on that issue:

> For now, I'll consider adding the same sort options that project views already support (by due
> date, priority, title, etc.) to the All Tasks view as a first step. The two-level sorting
> (primary + secondary key) is an interesting idea but adds significant UI complexity, I'll keep it
> in mind for a future iteration.

So this is parity with project views only — **no two-level sorting**, and no List/Board toggle
(board columns are sections, which has no obvious meaning across projects; happy to follow up if
you have a design in mind).

### Implementation notes

- **State lives in GSettings**, not the database, following the existing `today-sort-order` /
  `scheduled-sort-order` precedent — a filter view has no `Projects` row to hang columns off.
  Three new keys: `all-items-sort-order`, `all-items-sort-ascending`, `all-items-filters`.
  No schema migration is required.
- **The default is `project`, so existing behaviour is preserved** — the view still groups by
  project until the user chooses otherwise.
- Sorting by project *name* rather than raw `project_id` also fixes a small pre-existing mismatch:
  the list was ordered by UUID while `header_project_function` labelled the groups by name, so the
  grouping appeared arbitrary.

**Ordering within a project group.** Sorting by Project is a full ordering, not a project-name
comparison alone: project name, then project id, then priority (highest first, ties broken by
date added) within a group. Without the trailing keys the comparator returns `0` for any two tasks
in the same project, which leaves their relative order to the item load order and to `GtkListBox`'s
insertion for equal rows — arbitrary, and not stable across restarts. The project-id key also
keeps two projects that happen to share a name from interleaving under a single repeated header.
The ascending toggle flips the grouping only; priority stays highest-first in both directions, on
the reading that the toggle controls the grouping rather than the ranking inside it.

The comparator lives in `Util.set_item_project_sort_func` rather than inline in `Views.Filter`, so
it is unit-testable and available to the Label view if that follow-up lands. It delegates the
within-group comparison to the existing `Util.set_item_sort_func`, so priority semantics stay
identical to every other view. Tasks whose project row is gone collate under an empty name instead
of dereferencing null.
- Project headers are cleared under non-project sorts, where they would otherwise fragment the
  order into noise.
- The list is rebuilt on an idle callback rather than inline, because pagination selects which
  items load from the sorted backing list, and tearing rows down inside a signal emission is not
  safe. The idle source is cancelled in `clean_up ()`.
- The sort menu is gated behind a `sorting_supported` check so no other filter view is affected.
- The filters are applied to the backing list, not to the loaded rows. `add_items ()` pages that
  list `PAGE_SIZE` items at a time, so a filter evaluated only in the `GtkListBox` filter func can
  hide the whole first page while the matches sit behind it, unreachable without repeated
  *Load more*. The backing list now carries the filter the way it already carries the sort, so
  pagination pages over matching items and the *Load more* count means what it says.

### Testing instructions

1. Open **All Tasks** and use the new button in the header bar.
2. Each sort option reorders correctly, ascending and descending; the choice survives a restart.
3. Project headers appear only under the Project sort, and within each project the tasks run P1 to
   P4. Switching to descending reverses the project order while leaving priority highest-first.
   The order is identical after a restart.
4. Due-date, priority and label filters narrow the list; the indicator dot appears when the view is
   not in its default state; filters survive a restart.
5. Filter chips appear above the list, one per active due-date, priority or label filter, and are
   present immediately on restart. Removing a chip re-filters the list and clears the matching entry
   in the menu; removing the last one collapses the row.
6. With a filter active whose matches sort past the first page — e.g. *This Month* on a list of more
   than 30 tasks where the matching ones sort last — the matching tasks appear immediately rather
   than the list rendering empty, and *Load more* counts only matching tasks.
7. Regression check: Completed, Pinboard, Priority, Tomorrow, Anytime, Repeating and Unlabeled all
   share `Views.Filter`, and should be unchanged.

Manually tested on Fedora Asahi (aarch64) against a live Nextcloud CalDAV account, using the
Flatpak build from `build-aux/io.github.alainm23.planify.Devel.json`.

This also adds `test/core/`, covering the grouping comparator: priority ordering within a group, the
date-added tie-break, grouping under both sort directions, two projects sharing a name, orphaned
tasks, and that the comparator returns `0` only for an item compared with itself. It runs with
`meson test --suite core` alongside the existing `cli` and `caldav-integration` suites and needs no
database or display. The view code itself still has no harness and is covered manually. If you would
rather this PR stayed limited to the feature, I am happy to split the test target into its own.

Per `AI_POLICY.md`: this was developed with AI assistance. However, I have run and debugged the
result myself.

### DIFF
--- a/src/Views/Filter.vala
+++ b/src/Views/Filter.vala
@@ -32,11 +32,9 @@ public class Views.Filter : Adw.Bin {
     private const string SORT_ORDER_KEY = "all-items-sort-order";
     private const string SORT_ASCENDING_KEY = "all-items-sort-ascending";
     private const string FILTERS_KEY = "all-items-filters";
-    private const string SHOW_COMPLETED_KEY = "all-items-show-completed";
 
     private Widgets.ContextMenu.MenuPicker due_date_item;
     private Widgets.ContextMenu.MenuCheckPicker priority_filter;
-    private Widgets.ContextMenu.MenuSwitch show_completed_item;
     private Gtk.Revealer indicator_revealer;
     private uint rebuild_idle_id = 0;
 
@@ -78,15 +76,9 @@ public class Views.Filter : Adw.Bin {
             if (items_list == null || items_list.size == 0) return false;
             if (selected_project_ids.size == 0 && filter.filters.size == 0) return true;
             foreach (var item in items_list) {
-                if (selected_project_ids.size > 0 && !selected_project_ids.contains (item.project_id)) {
-                    continue;
+                if (item_matches_filters (item)) {
+                    return true;
                 }
-
-                if (sorting_supported && !Utils.TaskUtils.items_filter_func (item, filter.filters)) {
-                    continue;
-                }
-
-                return true;
             }
             return false;
         }
@@ -148,8 +140,7 @@ public class Views.Filter : Adw.Bin {
         project_picker_core.multiselect_changed.connect ((ids) => {
             selected_project_ids = ids;
             update_project_filter_label ();
-            listbox.invalidate_filter ();
-            validate_placeholder ();
+            rebuild_list ();
         });
 
         project_filter_revealer = new Gtk.Revealer () {
@@ -181,7 +172,6 @@ public class Views.Filter : Adw.Bin {
             halign = END,
             valign = START,
             sensitive = false,
-            reveal_child = false
         };
 
         var view_setting_overlay = new Gtk.Overlay () {
@@ -296,17 +286,7 @@ public class Views.Filter : Adw.Bin {
         toolbar_view.add_top_bar (headerbar);
 
         listbox.set_filter_func ((row) => {
-            var item = ((Layouts.ItemRow) row).item;
-
-            if (selected_project_ids.size > 0 && !selected_project_ids.contains (item.project_id)) {
-                return false;
-            }
-
-            if (sorting_supported && !Utils.TaskUtils.items_filter_func (item, filter.filters)) {
-                return false;
-            }
-
-            return true;
+            return item_matches_filters (((Layouts.ItemRow) row).item);
         });
 
         child = toolbar_view;
@@ -359,12 +339,12 @@ public class Views.Filter : Adw.Bin {
 
         if (sorting_supported) {
             signal_map[Services.Settings.get_default ().settings.changed[SORT_ORDER_KEY].connect (() => {
-                apply_sort_changed ();
+                rebuild_list ();
                 check_default_filters ();
             })] = Services.Settings.get_default ().settings;
 
             signal_map[Services.Settings.get_default ().settings.changed[SORT_ASCENDING_KEY].connect (() => {
-                apply_sort_changed ();
+                rebuild_list ();
                 check_default_filters ();
             })] = Services.Settings.get_default ().settings;
 
@@ -392,17 +372,17 @@ public class Views.Filter : Adw.Bin {
 
     private void apply_filters_changed () {
         save_filters ();
-        listbox.invalidate_filter ();
-        validate_placeholder ();
+        rebuild_list ();
         check_default_filters ();
     }
 
     /**
-     * Re-sorts and re-pages the list after the sort settings change. The whole list
-     * has to be rebuilt rather than merely invalidated, because pagination decides
-     * which items are loaded at all from the sorted backing list.
+     * Re-sorts, re-filters and re-pages the list after the sort or filter settings change.
+     * The whole list has to be rebuilt rather than merely invalidated, because pagination
+     * decides which items are loaded at all from the sorted and filtered backing list —
+     * invalidating the loaded rows alone would leave matches beyond the first page unreachable.
      */
-    private void apply_sort_changed () {
+    private void rebuild_list () {
         // Deferred to an idle callback rather than run inline. Rebuilding tears down and
         // frees every ItemRow, and this runs from inside a signal emission (a GSettings
         // "changed" handler) that can still touch those rows once we return — which shows
@@ -479,6 +459,22 @@ public class Views.Filter : Adw.Bin {
         project_filter_revealer.reveal_child = filter is Objects.Filters.Completed;
     }
 
+    /**
+     * Whether an item survives the view's active filters — the project picker of the
+     * Completed view and the sort/filter menu of the All Tasks view.
+     */
+    private bool item_matches_filters (Objects.Item item) {
+        if (selected_project_ids.size > 0 && !selected_project_ids.contains (item.project_id)) {
+            return false;
+        }
+
+        if (sorting_supported && !Utils.TaskUtils.items_filter_func (item, filter.filters)) {
+            return false;
+        }
+
+        return true;
+    }
+
     private void add_items () {
         foreach (Layouts.ItemRow row in items.values) {
             row.clean_up ();
@@ -526,13 +522,20 @@ public class Views.Filter : Adw.Bin {
             foreach (Objects.Item item in Services.Store.instance ().get_items_no_parent (false)) {
                 items_list.add (item);
             }
+        }
 
-            if (show_completed ()) {
-                foreach (Objects.Item item in Services.Store.instance ().get_items_no_parent (true)) {
-                    items_list.add (item);
-                }
+        // The filters have to be applied to the backing list, not merely to the loaded rows:
+        // pagination decides which items are loaded at all, so filtering afterwards would
+        // hide the first page and never reach the matches behind it. Keeping the backing
+        // list filtered also keeps the "Load more" count honest.
+        var filtered_list = new Gee.ArrayList<Objects.Item> ();
+        foreach (Objects.Item item in items_list) {
+            if (item_matches_filters (item)) {
+                filtered_list.add (item);
             }
         }
+
+        items_list = filtered_list;
 
         if (filter is Objects.Filters.Completed) {
             items_list.sort ((a, b) => {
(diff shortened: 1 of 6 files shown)


CR0CKER:
Pushed an amendment. Sorting by Project previously compared project names only, so two tasks in
the same project compared equal and their order fell through to the item load order — arbitrary,
and not stable across restarts. The comparator now carries project id and priority as further
keys, so tasks read P1 → P4 inside each group and two projects sharing a name cannot interleave
under one header.

I moved the comparator into `Util.set_item_project_sort_func` so it could be unit-tested, and added
a small `test/core/` suite for it. Happy to split that test target into its own PR if you would
rather keep this one to the feature.

CR0CKER:
One more addition: the All Tasks view now shows the same removable filter chips as the project,
Today and Scheduled views. It reuses `Widgets.FilterFlowBox` as-is — the filters were already in the
`BaseObject` bag the widget reads, so this is only the missing wiring, and removing a chip goes
through the same `remove_filter ()` path the menu does.

src/Views/Filter.vala
```diff
@@ -709,6 +890,364 @@ public class Views.Filter : Adw.Bin {
...
+
+    private bool show_completed () {
+        return sorting_supported && Services.Settings.get_default ().settings.get_boolean (SHOW_COMPLETED_KEY);
+    }
+
+    /**
+     * Reveals the dot on the view-settings button whenever the view is not showing
+     * its default sort and no filters, mirroring the project view's affordance.
+     */
+    private void check_default_filters () {
```
alainm23, line 1186: `indicator_revealer` is always initialised in `construct`, so the null guard in `check_default_filters()` is unnecessary and inconsistent with the rest of the widget code. Safe to remove.

src/Views/Filter.vala
```diff
@@ -709,6 +890,364 @@ public class Views.Filter : Adw.Bin {
...
+        sorted_by_item.add_item (_("Due Date"), SortedByType.DUE_DATE.to_string ());
+        sorted_by_item.add_item (_("Date Added"), SortedByType.ADDED_DATE.to_string ());
+        sorted_by_item.add_item (_("Date Modified"), SortedByType.UPDATED_DATE.to_string ());
+        sorted_by_item.add_item (_("Priority"), SortedByType.PRIORITY.to_string ());
+
+        var sort_order_item = new Widgets.ContextMenu.MenuSwitch (_("Ascending Order"), "vertical-arrows-long-symbolic") {
+            active = Services.Settings.get_default ().settings.get_boolean (SORT_ASCENDING_KEY)
+        };
+
+        due_date_item = new Widgets.ContextMenu.MenuPicker (_("Duedate"), "month-symbolic") {
```
alainm23, line 911: When a filter is active (e.g. "This Month"), items that match but fall outside the first loaded page are never shown, because pagination stops before reaching them. The backing list needs to either be fully loaded when filters are active, or the "Load more" button should keep loading until at least one visible item is found.

src/Views/Filter.vala
```diff
@@ -709,6 +890,364 @@ public class Views.Filter : Adw.Bin {
...
+        });
+
+        priority_filter = new Widgets.ContextMenu.MenuCheckPicker (_("Priority"), "flag-outline-thick-symbolic");
+        priority_filter.set_items (priority_items);
+
+        var labels_filter = new Widgets.ContextMenu.MenuItem (_("Filter by Labels"), "tag-outline-symbolic") {
+            arrow = true
+        };
+
+        show_completed_item = new Widgets.ContextMenu.MenuSwitch (_("Show Completed"), "check-round-outline-symbolic") {
```
alainm23: `show_completed` triggers a full list rebuild via `apply_sort_changed()`. A better approach would be to reuse `Dialogs.CompletedTasks` as a panel below the list, but it currently only supports per-project scope. Worth extending it in a follow-up PR.

src/Views/Filter.vala
```diff
@@ -129,10 +167,32 @@ public class Views.Filter : Adw.Bin {
...
+            css_classes = { "indicator" }
+        };
+
+        indicator_revealer = new Gtk.Revealer () {
             transition_type = Gtk.RevealerTransitionType.CROSSFADE,
+            child = indicator_grid,
+            halign = END,
+            valign = START,
+            sensitive = false,
+            reveal_child = false
```
alainm23:
```suggestion
```

src/Views/Filter.vala
```diff
@@ -129,10 +167,32 @@ public class Views.Filter : Adw.Bin {
...
+            margin_end = 6,
+            css_classes = { "indicator" }
+        };
+
+        indicator_revealer = new Gtk.Revealer () {
             transition_type = Gtk.RevealerTransitionType.CROSSFADE,
+            child = indicator_grid,
+            halign = END,
+            valign = START,
+            sensitive = false,
```
alainm23, line 174:
```suggestion
```

src/Views/Filter.vala
```diff
@@ -129,10 +167,32 @@ public class Views.Filter : Adw.Bin {
...
+            margin_top = 6,
+            margin_end = 6,
+            css_classes = { "indicator" }
+        };
+
+        indicator_revealer = new Gtk.Revealer () {
             transition_type = Gtk.RevealerTransitionType.CROSSFADE,
+            child = indicator_grid,
+            halign = END,
+            valign = START,
```
alainm23, line 173:
```suggestion
            valign = START
```

alainm23 (changes requested):

alainm23:
Hey, great work on this! The implementation is clean and well thought out. 

Regarding the Show Completed option — I think we could remove it from this PR for now. The right approach would be to reuse `Dialogs.CompletedTasks` for this, but it currently only supports per-project scope so it would need to be extended first. That feels like the scope of a separate PR.

Overall, looks good to me! 🙌

CR0CKER:
Thanks for the review — all three are addressed in the push above.

**Pagination vs. filters.** You are right, and it was worse than a missed page: `has_visible_items` scans the whole backing list, so the placeholder did not appear either — the view rendered a blank list with a `+30 tasks` button under it. Rather than loading everything when filters are active, or looping *Load more* until something shows, I applied the filter to the backing list itself, which is the same move the sort already makes in `add_items ()` for the same reason. Pagination then pages over matching items, and the *Load more* count stops counting tasks the user cannot see. The shared predicate is now `item_matches_filters ()`, reused by the listbox filter func, `has_visible_items` and the live `item_added` path, and filter changes go through the existing idle-coalesced rebuild (`apply_sort_changed ()` renamed to `rebuild_list ()`, since it is no longer only about sorting).

Worth noting: this also closes the same hole for the Completed view's project picker, which has been able to page past matches since before this PR — `selected_project_ids` was evaluated in the listbox filter func only. That is not a new regression from this branch, but it is the same bug and it is fixed by the same change.

**Show Completed.** Removed — the feature, the `all-items-show-completed` key, and its testing step. Agreed that `Dialogs.CompletedTasks` is the right home for it once it takes a scope other than a single project; I would rather leave that to its own PR than half-extend it here.

**The null guard.** Removed. You are right that nothing reaches `check_default_filters ()` before `construct` assigns the revealer.

**The revealer initialiser.** I took `reveal_child = false` out — it is the `GtkRevealer` default and none of the sibling views set it. I kept `sensitive = false,` though: `Views.Today`, `Views.Scheduled` and `Views.Project` all set it on their indicator revealer, and it is what makes the 9x9 dot transparent to input — `gtk_widget_pick` skips insensitive widgets by default, so without it the dot sits in the overlay and swallows clicks on that corner of the menu button. Happy to drop it too if you would rather, but I think it is load-bearing.

Issue #1220, opened 2024-03-28
Jump to list via ctrl/alt + number keyboard shortcuts


oktayacikalin:
**Is your feature request related to a problem? Please describe.**
On my desktop, I can jump to my favorites by using meta+number. In Firefox I can jump to tabs using alt+number. This is great.

**Describe the solution you'd like**
I would like to jump to my lists using ctrl+number.

**Describe alternatives you've considered**
Use ctrl+something to select the list of list and the use the cursor keys to navigate around. But ctrl+number is much faster.

**Additional context**
Why does Firefox use alt+number instead of ctrl+number?!

alainm23:
How could the number of the list be defined? Would it be configurable?

oktayacikalin:
I would just assign a number to every list from top to bottom. So the order of the lists decides.

CR0CKER:
I've opened #2656 implementing this.

On your question about how the number of the list would be defined — this takes @oktayacikalin's
suggestion from above: **the sidebar order defines the number**, top to bottom, across source
groups. That means no configuration, no new settings and no extra UI, and the numbering follows
drag-reordering and the Alphabetically sort option automatically, because it is derived from exactly
what the sidebar renders rather than stored separately.

- `Ctrl+1`–`Ctrl+8` open that position; `Ctrl+9` opens the last project, whatever the count —
  the same convention browsers and chat apps use for numbered tabs.
- `Ctrl+Page Down` / `Ctrl+Page Up` cycle to the next and previous project, which covers the case
  where you have more lists than digits.

I used `Ctrl`+number rather than `Alt`+number to match the app's existing `Ctrl`+letter view
shortcuts, and because the HIG reserves `Alt`+arrows for navigation and discourages `Alt` for
application shortcuts.

If you would rather the numbers were user-assignable, that could be added later on top of these
bindings without changing them.

PR #2656, opened 2026-08-30
feat: keyboard shortcuts to jump to and cycle through projects

CR0CKER:
### What changed

Projects can now be reached from the keyboard:

- **`Ctrl+1`–`Ctrl+8`** open the project at that position in the sidebar; **`Ctrl+9`** opens the
  last one, whatever the count.
- **`Ctrl+Page Down` / `Ctrl+Page Up`** move to the next and previous project, wrapping at both
  ends. From a non-project view they open the first or last project.

### Why

Six shortcuts already jump to a view (`Ctrl+I` Inbox, `Ctrl+T` Today, `Ctrl+U` Scheduled, `Ctrl+L`
Labels, `Ctrl+P` Pinboard, `Ctrl+H` Home), but there was no way to reach a **project** without the
mouse — the sidebar has no key handling at all. This closes that gap and is the request in #1220.

### How the numbering is defined

#1220 asked whether list numbers would be configurable. This implementation takes the approach the
reporter suggested there — **the sidebar order defines the number** — which needs no configuration,
no new settings and no UI:

- `Store.get_projects_display_order ()` reproduces exactly what the sidebar renders: visible sources
  by `child_order`, then each source's top-level, non-archived projects, honouring the
  Alphabetically sort setting. Numbering therefore always matches what the user sees, and follows
  drag-reordering automatically.
- Subprojects are excluded, since they render nested inside their parent rather than as siblings.
- The list is recomputed on each keypress, so adding, archiving, reordering or hiding a source takes
  effect immediately without a restart.

**The designated inbox is excluded from the numbering.** `Layouts.SidebarSourceRow.add_row_project
()` builds a row for the inbox project like any other, and `Layouts.ProjectRow` then keeps that row
hidden behind its revealer (`main_revealer.reveal_child = !project.is_inbox_project`). Counting the
rows the sidebar *builds* rather than the ones it *shows* therefore puts a position under a list the
user cannot see, and every list after the inbox comes out off by one. `Ctrl+I` already opens the
Inbox, so excluding it makes nothing unreachable. Because `is_inbox_project` is derived from the
`local-inbox-project-id` setting and the list is rebuilt on each keypress, changing which list is
the default is picked up immediately, with no cache to invalidate.

One consequence worth stating: while viewing the Inbox, `Ctrl+Page_Down` opens the first project and
`Ctrl+Page_Up` the last, rather than continuing from the inbox's old position. The inbox sits
outside the cycle in the same way it sits outside the sidebar list.

If you would prefer numbers to be user-assignable, that could be layered on later without changing
these bindings.

### Why these keys

- **`Ctrl`+number** is the jump-to-nth convention in Chrome, Slack, Discord, VS Code and Obsidian,
  and it matches Planify's own `Ctrl`+letter view shortcuts. **`Ctrl+9` = last** follows the same
  convention browsers use.
- **`Ctrl+Page Down`/`Up`** is the GNOME next/previous convention (Terminal, Nautilus, Text Editor).
- `Alt`+number was avoided because the HIG reserves `Alt+←/→/↑` for navigation and discourages `Alt`
  for app shortcuts; `Super` belongs to the shell; `Ctrl+Tab` is GTK focus traversal.

None of these combinations were previously bound anywhere in the app.

### Implementation notes

- The numbered jump is a **single parameterised action** registered by detailed action name
  (`win.action_view_project(1)`), rather than nine near-identical actions. Because the existing
  enable/disable loops in `ActionManager` only prepend `ACTION_PREFIX`, this needed no changes to
  that machinery.
- All new accelerators are `Ctrl` chords and so live in `action_accelerators`, not
  `typing_accelerators` — they are safe while typing.
- Navigation reuses `pane_selected (PaneType.PROJECT, id)`, the same call `ProjectRow` makes on
  click, so the sidebar highlight updates reactively with no extra work.
- Added a small `MainWindow.get_current_project ()` accessor, needed to know where to cycle from.
- The three bindings are added to the shortcuts window (`shortcuts.ui`) so they are discoverable.

### Testing instructions

1. `Ctrl+1`–`Ctrl+8` open the 1st–8th project top to bottom across source groups; `Ctrl+9` opens the
   last. Numbers with no project do nothing.
2. `Ctrl+Page Down`/`Up` cycle both ways and wrap; from Today or Labels they jump to the first or
   last project.
3. The Inbox is skipped: with an inbox present, `Ctrl+1` opens the first list shown in the sidebar
   and cycling never lands on the Inbox. `Ctrl+I` still opens it. Designating a different list as
   the default in Preferences moves the exclusion to that list immediately.
4. Switch Preferences to sort projects Alphabetically — the numbering follows the re-ordered sidebar.
5. Reorder by drag, add, archive a project, or hide a source — numbering updates without restart.
6. Digits typed in a task title or Quick Add do not trigger navigation.
7. `F1` lists the new shortcuts; the existing 15 still work.

Tested on Fedora Asahi (aarch64) with a Nextcloud CalDAV account.

Per `AI_POLICY.md`: this was developed with AI assistance. However, I have run and debugged the
result myself.

### DIFF
--- a/core/Services/Store.vala
+++ b/core/Services/Store.vala
@@ -315,15 +315,22 @@ public class Services.Store : GLib.Object {
 
     /**
      * Returns every project in the order the sidebar displays it: visible sources by child_order,
-     * and within each source its top-level, non-archived projects.
+     * and within each source its top-level, non-archived projects, again by child_order.
+     *
+     * Both orderings have to be computed rather than read off the store's own list: that list is
+     * ordered once at load, while reordering a project by dragging it writes the new child_order
+     * onto the live Objects.Project and to the database without moving anything in the list. Custom
+     * order taken from the list order would therefore stay on the pre-drag order until the next
+     * restart, while the sidebar showed the new one.
      *
      * Matches what the sidebar actually shows, so callers that address projects by position agree
      * with what the user sees. That is two conditions, not one: Layouts.SidebarSourceRow's
      * add_row_project () decides which rows are built, and Layouts.ProjectRow then keeps the
      * designated inbox hidden (main_revealer.reveal_child = !project.is_inbox_project), so a row
      * existing in the sidebar's list box does not mean it is visible. The alphabetical mode is the
-     * same as projects_sort_func (). Subprojects are excluded because they are rendered nested
-     * inside their parent, not as siblings in this list.
+     * same as projects_sort_func (); custom order ignores the projects-ordered direction exactly as
+     * the sidebar does, which installs no sort function at all in that mode. Subprojects are
+     * excluded because they are rendered nested inside their parent, not as siblings in this list.
      */
     public Gee.ArrayList<Objects.Project> get_projects_display_order () {
         Gee.ArrayList<Objects.Project> return_value = new Gee.ArrayList<Objects.Project> ();
@@ -354,6 +361,10 @@ public class Services.Store : GLib.Object {
             if (alphabetically) {
                 source_projects.sort ((a, b) => {
                     return ordered == 0 ? b.name.collate (a.name) : a.name.collate (b.name);
+                });
+            } else {
+                source_projects.sort ((a, b) => {
+                    return a.child_order - b.child_order;
                 });
             }
 
--- a/core/Services/Store.vala
+++ b/core/Services/Store.vala
@@ -313,6 +313,56 @@ public class Services.Store : GLib.Object {
         }
     }
 
+    /**
+     * Returns every project in the order the sidebar displays it: visible sources by child_order,
+     * and within each source its top-level, non-archived projects.
+     *
+     * Matches what the sidebar actually shows, so callers that address projects by position agree
+     * with what the user sees. That is two conditions, not one: Layouts.SidebarSourceRow's
+     * add_row_project () decides which rows are built, and Layouts.ProjectRow then keeps the
+     * designated inbox hidden (main_revealer.reveal_child = !project.is_inbox_project), so a row
+     * existing in the sidebar's list box does not mean it is visible. The alphabetical mode is the
+     * same as projects_sort_func (). Subprojects are excluded because they are rendered nested
+     * inside their parent, not as siblings in this list.
+     */
+    public Gee.ArrayList<Objects.Project> get_projects_display_order () {
+        Gee.ArrayList<Objects.Project> return_value = new Gee.ArrayList<Objects.Project> ();
+
+        var visible_sources = new Gee.ArrayList<Objects.Source> ();
+        foreach (var source in sources) {
+            if (source.is_visible) {
+                visible_sources.add (source);
+            }
+        }
+
+        visible_sources.sort ((a, b) => {
+            return a.child_order - b.child_order;
+        });
+
+        bool alphabetically = Services.Settings.get_default ().settings.get_enum ("projects-sort-by") == 1;
+        int ordered = Services.Settings.get_default ().settings.get_enum ("projects-ordered");
+
+        foreach (var source in visible_sources) {
+            var source_projects = new Gee.ArrayList<Objects.Project> ();
+
+            foreach (var project in get_projects_by_source (source.id)) {
+                if (project.parent_id == "" && !project.is_archived && !project.is_inbox_project) {
+                    source_projects.add (project);
+                }
+            }
+
+            if (alphabetically) {
+                source_projects.sort ((a, b) => {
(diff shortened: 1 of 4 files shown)


CR0CKER:
Pushed an amendment. The numbering was counting the designated inbox, which `ProjectRow` hides
behind its revealer, so with an inbox present every position after it was off by one — `Ctrl+4`
would open the Inbox rather than the 4th list in the sidebar. The ordering now filters on
`is_inbox_project` as well, matching what the sidebar actually shows. `Ctrl+I` already opens the
Inbox, so nothing became unreachable.

core/Services/Store.vala
```diff
@@ -313,6 +313,56 @@ public class Services.Store : GLib.Object {
...
+     *
+     * Matches what the sidebar actually shows, so callers that address projects by position agree
+     * with what the user sees. That is two conditions, not one: Layouts.SidebarSourceRow's
+     * add_row_project () decides which rows are built, and Layouts.ProjectRow then keeps the
+     * designated inbox hidden (main_revealer.reveal_child = !project.is_inbox_project), so a row
+     * existing in the sidebar's list box does not mean it is visible. The alphabetical mode is the
+     * same as projects_sort_func (). Subprojects are excluded because they are rendered nested
+     * inside their parent, not as siblings in this list.
+     */
+    public Gee.ArrayList<Objects.Project> get_projects_display_order () {
```
alainm23, line 335: Nice feature, I've been wanting this one. Tried it out and hit a snag though: if I reorder my projects by dragging them around, the number shortcuts and Ctrl+Page Down/Up don't follow the new order — they stick to the old one until I restart the app. Sorting alphabetically works fine, it's just the default custom order.

Poking around, it looks like `get_projects_display_order ()` doesn't sort by `child_order` in the custom-order case, so it ends up trusting whatever order the projects happen to be in memory, which only gets sorted on startup. Dragging updates the DB but not that in-memory list.

Sorting the same way there should do it:
```
} else {
    source_projects.sort ((a, b) => a.child_order - b.child_order);
}
```

data/resources/ui/shortcuts.ui
```diff
@@ -129,6 +129,27 @@
                 <property name="title" translatable="yes" context="shortcut window">Open Pinboard</property>
               </object>
             </child>
+
+            <child>
+              <object class="GtkShortcutsShortcut" id="project-number">
+                <property name="accelerator">&lt;ctrl&gt;1...&lt;ctrl&gt;9</property>
```
alainm23, line 135: In shortcuts.ui, the accelerator <ctrl>1...<ctrl>9 uses ... as a range, but GTK doesn't parse that — it renders literally. Probably intentional to show a readable range, just worth confirming it looks right in the shortcuts window and not like a broken accelerator.

Image

alainm23 (changes requested):

CR0CKER:
Good catch, thank you! Fixed in the push above with the sort you suggested.

For the record of why it only bit the custom order: `update_projects_child_order ()` writes the new
`child_order` onto the live `Objects.Project` and to the database, but nothing reorders the store's
in-memory list, which is ordered once at load. In alphabetical mode that did not matter, because
`get_projects_display_order ()` sorted by name anyway; in custom mode it fell through unsorted and
so read the load-time order, while the sidebar — which installs no sort function in that mode — had
just moved the row itself. Sidebar right, shortcuts stale, until the next restart re-read the
database. Sorting by `child_order` there works because the drag has already updated it on the
objects themselves, so the shortcuts now follow the drag immediately.

One intentional omission: the custom branch does not apply the `projects-ordered` direction. The
sidebar consults that setting only inside `projects_sort_func ()`, which it installs only when
sorting alphabetically, so applying it here would make the shortcuts disagree with the sidebar in
exactly the case this helper exists to match.

No unit test for it, I am afraid — `get_projects_display_order ()` reads `Services.Store.instance ()`
and GSettings, so covering it needs a database and a settings backend, where the `test/core/` suite
covers pure functions. It is verified by hand: drag a project, then `Ctrl+N` and
`Ctrl+Page Down`/`Up` follow the new order without a restart, in both the custom and alphabetical
modes and with the sort switched by the drag itself.

**On the accelerator** — that one is intentional and, I think, correct as rendered. `...` is
documented `GtkShortcutsShortcut:accelerator` syntax for a range ("it is also possible to specify
ranges of shortcuts, using `...` between the keys"), and the property's own example in the GTK docs
is `<alt>1...<alt>9`, the same shape. Your screenshot is `GtkShortcutLabel` doing the right thing
with it: two real key boxes with an ellipsis between them, rather than the raw string. If you would
rather it read differently in the shortcuts window I am happy to change it, but it is parsed, not
literal.

PR #2670, opened 2026-09-01
fix: glib critical for unparseable JSON in Utils.JsonUtils

byquanton:


byquanton:
I think this makes #2647 obsolete

alainm23 (approved):

PR #2679, opened 2026-09-08
Add backup retention limit and paginate the backup list

### LINKED ISSUE #2632: Handling the growing number of backups
### Prerequisites

- [x] I have searched the [existing issues](https://github.com/alainm23/planify/issues) and this feature has not been requested yet

### Is your feature request related to a problem?

Hi, the backup function is working, for which I’m very grateful.

Something I’ve noticed is that the list of backups is getting longer and longer.
There’s no way to automatically limit the number of backups created and stored by Planify – for example, by setting a rule.
Manual deletion is, of course, possible, but as far as I can see, only one backup at a time and not by selecting multiple items.
The longer the list gets, the more I’ll need to keep an eye on deleting them regularly, so it doesn’t become exhausting at some point.

My question: How could Planify solve this elegantly?

### Describe the solution you'd like

Perhaps my enquiry already includes the first few ideas :)

### Alternatives you've considered

_No response_

### Additional context

_No response_

alainm23:
Backups previously accumulated forever — a new one was created every day and none were ever removed. This adds an automatic retention limit and a cleaner backup list.

A new "Backups to Keep" setting (default 30, zero keeps all) prunes the oldest backups after each new one is created, and applies immediately when lowered. The backup list is now paginated with numbered page buttons, the
current page highlighted, replacing the old "only the last 30 shown" note. The backup location is shown as a proper row with an Open button instead of a raw inline path.

Close: #2632

image

image


4jNsY6fCVqZv:
@alainm23 Thank you!
Have you also thought about making it possible to select multiple backups at once for manual deletion?
Then, for example, I could keep one backup per week or per month and easily delete all the backups in between.

alainm23:
> @alainm23 Thank you! Have you also thought about making it possible to select multiple backups at once for manual deletion? Then, for example, I could keep one backup per week or per month and easily delete all the backups in between.

Good idea! Multi-select deletion isn't in this change, but it makes sense for exactly that keep-one-per-week/month cleanup. I'll keep it in mind for a later improvement.

